PR #1, opened 2015-02-09
Remove whitespace errors, reorder alphabetically

mnagy:


Dockerfile.rhel7
```diff
@@ -1,27 +1,27 @@
...
-    sqlite-devel && \
+    mysql-libs \
+    openssl-devel \
+    postgresql-devel \
+    procps-ng \
+    sqlite-devel \
+    tar \
+    unzip \
+    wget \
+    which \
```
mfojtik: why is 'which' here twice?

Dockerfile
```diff
@@ -1,28 +1,28 @@
...
+    mysql-libs \
+    openssl-devel \
     postgresql-devel \
-    sqlite-devel && \
+    procps-ng \
+    sqlite-devel \
+    tar \
+    unzip \
+    wget \
+    which \
```
mfojtik: same here

mnagy: 'cause I didn't run it through uniq :) Fixed.

PR #4, opened 2015-02-16
Minor repository tweaks

mfojtik:


mfojtik:
@bparees @jhadvig fyi

I think the naming convention for the base image should be: `openshift/base-OSVERSION`

bparees:
works for me, but we need to fix all the images referencing this, right?

mfojtik:
there are none atm.

Dockerfile.rhel7
```diff
@@ -1,5 +1,11 @@
 FROM rhel7
 
+# This image is an base image for OpenShift v3 Docker images.
```
soltysh: Suggestion: This is base image for OpenShift...

Dockerfile
```diff
@@ -1,5 +1,12 @@
 FROM centos:centos7
 
+# This image is an base image for OpenShift v3 Docker images.
```
soltysh: Suggestion: This is base image for OpenShift...

Dockerfile
```diff
@@ -1,5 +1,12 @@
...
+# It includes all dependencies that users might need to debug broken
+# applications or services as well as common build dependencies for all
+# languages and frameworks.
+
+MAINTAINER Jakub Hadvig <jhadvig@redhat.com>
```
soltysh: The best for this would be to have a devexp alias for that purpose, I guess.

Dockerfile
bparees: This is *a* base image for...... or *the* base image for.

Dockerfile.rhel7
bparees: this is a/the base image for

Dockerfile
soltysh: Yeah, the articles a, an and the are always a problem for me :wink:

Dockerfile
jhadvig: @soltysh this looks like the right approach but I doubt we have something like that.
@mfojtik thoughts?

PR #5, opened 2015-02-18
WIP: Add more common functionality to common image

mfojtik:


mfojtik:
@soltysh @jhadvig review pls, see the referenced PR for more infos.

Dockerfile
```diff
@@ -1,5 +1,8 @@
 FROM centos:centos7
 
+# openshift/base-centos7
+MAINTAINER Jakub Hadvig <jhadvig@redhat.com>
```
jhadvig: @mfojtik this questio was already risen in the https://github.com/openshift/sti-base/pull/4/files 
Like I'm fine to be the MAINTAINER but I dont think that there should be just one person. And since there can't be multiple maintainers we have two options. Either dont put the directive into the Dockerfiles at all, or create something similar: https://github.com/docker/docker/blob/master/MAINTAINERS

Thoughts?

bparees: i don't think people figuring out how to get a hold of someone in openshift is going to be a huge problem :)

Dockerfile
```diff
@@ -25,3 +28,32 @@ RUN yum install -y --setopt=tsflags=nodocs \
...
+ENV STI_SCRIPTS_URL image:///usr/local/sti
+
+# Setup the 'openshift' user that is used for the build execution and for the
+# application runtime execution.
+# TODO: Use better UID and GID values
+RUN mkdir -p /opt/openshift/src && \
+    groupadd -r openshift -f -g 433 && \
+    useradd -u 431 -r -g openshift -d /opt/openshift/src -s /sbin/nologin \
+            -c "OpenShift Application User" openshift && \
+    chown -R openshift:openshift /opt/openshift
```
jhadvig: How about:
- sti
- stiuser
- sti-user

any other ?

Dockerfile
```diff
@@ -25,3 +28,32 @@ RUN yum install -y --setopt=tsflags=nodocs \
...
+
+# Directory with the sources is set as the working directory so all STI scripts
+# can execute relative to this path
+WORKDIR /opt/openshift/src
+
+# Copy the STI scripts from the builder repository to the image
+# This instruction is executed when building the language builder image.
+ONBUILD COPY ./.sti/bin/ /usr/local/sti
+
+CMD ["usage"]
```
jhadvig: Since the sti `usage` script wont be part of the base image I doubt this will work. Or ?

Dockerfile
bparees: @smarterclayton weighed in on the email thread suggesting "default" which is used in other docker images.

Dockerfile
```diff
@@ -25,3 +28,32 @@ RUN yum install -y --setopt=tsflags=nodocs \
...
+# application runtime execution.
+# TODO: Use better UID and GID values
+RUN mkdir -p /opt/openshift/src && \
+    groupadd -r openshift -f -g 433 && \
+    useradd -u 431 -r -g openshift -d /opt/openshift/src -s /sbin/nologin \
+            -c "OpenShift Application User" openshift && \
+    chown -R openshift:openshift /opt/openshift
+
+# The $HOME is not set by default, but some applications needs this variable
+ENV HOME /opt/openshift/src
```
jhadvig: Did you consider to replace the `openshift` with something else? 
eg: sti
Just wondering whether he have the right name here.

Dockerfile
smarterclayton:
```
$ docker run centos:7 cat /etc/passwd
root:x:0:0:root:/root:/bin/bash
bin:x:1:1:bin:/bin:/sbin/nologin
daemon:x:2:2:daemon:/sbin:/sbin/nologin
adm:x:3:4:adm:/var/adm:/sbin/nologin
lp:x:4:7:lp:/var/spool/lpd:/sbin/nologin
sync:x:5:0:sync:/sbin:/bin/sync
shutdown:x:6:0:shutdown:/sbin:/sbin/shutdown
halt:x:7:0:halt:/sbin:/sbin/halt
mail:x:8:12:mail:/var/spool/mail:/sbin/nologin
operator:x:11:0:operator:/root:/sbin/nologin
games:x:12:100:games:/usr/games:/sbin/nologin
ftp:x:14:50:FTP User:/var/ftp:/sbin/nologin
nobody:x:99:99:Nobody:/:/sbin/nologin
```

```
$ docker run fedora cat /etc/passwd
root:x:0:0:root:/root:/bin/bash
bin:x:1:1:bin:/bin:/sbin/nologin
daemon:x:2:2:daemon:/sbin:/sbin/nologin
adm:x:3:4:adm:/var/adm:/sbin/nologin
lp:x:4:7:lp:/var/spool/lpd:/sbin/nologin
sync:x:5:0:sync:/sbin:/bin/sync
shutdown:x:6:0:shutdown:/sbin:/sbin/shutdown
halt:x:7:0:halt:/sbin:/sbin/halt
mail:x:8:12:mail:/var/spool/mail:/sbin/nologin
operator:x:11:0:operator:/root:/sbin/nologin
games:x:12:100:games:/usr/games:/sbin/nologin
ftp:x:14:50:FTP User:/var/ftp:/sbin/nologin
nobody:x:99:99:Nobody:/:/sbin/nologin
systemd-timesync:x:999:998:systemd Time Synchronization:/:/sbin/nologin
systemd-network:x:998:997:systemd Network Management:/:/sbin/nologin
systemd-resolve:x:997:996:systemd Resolver:/:/sbin/nologin
systemd-bus-proxy:x:996:995:systemd Bus Proxy:/:/sbin/nologin
dbus:x:81:81:System message bus:/:/sbin/nologin
```

Our name should fit naturally in this list.  I veto "sti" "openshift".  I would be in favor of "build" or "builder" or "default" or "insert generic here"

smarterclayton: Or "user"

soltysh: +1 for user

Dockerfile
soltysh: I'd propose adding something along these lines, so that running that image should print user some useful information about the purpose of this image.

Dockerfile.rhel7
```diff
@@ -1,5 +1,8 @@
 FROM rhel7
```
soltysh: Isn't there `rhel7:rhel7` or something similar?

Dockerfile
soltysh: And the dir should be `/opt/app/`

mfojtik: /opt/app is too generic.

Dockerfile
mfojtik: @jhadvig I was asking @danmcp and he suggested to put a real person name and email there. So congrats for being and official maintainer :D

Dockerfile
mfojtik: I don't like 'user' as it makes the USER look weird:

```
USER user
```

I like 'default', will update the Dockerfile with that. Thanks guys for suggestions

Dockerfile
mfojtik: I think this should be solved in this way:

```
ONBUILD CMD ["usage"]
CMD ["/usr/local/sti/base-usage"]
```

Dockerfile
mfojtik: This place is where the application source code is stored. It has nothing to do with STI. I would rather to keep this platform specific than making it 'STI'.

Dockerfile.rhel7
mfojtik: dunno, @jhadvig ?

Dockerfile
mfojtik: @soltysh @smarterclayton also I think we should not be abusing `/opt` but we should be installing the sources into `/usr/src' instead. This is what upstream is doing and all Docker official images are using that path.

mfojtik:
@smarterclayton @soltysh @jhadvig @bparees all comments should be addressed now.

Dockerfile.rhel7
jhadvig: @soltysh no there isnt. I would keep it rhel7 and maybe put he centos Dockerfile to just `centos7` so will have the latest

soltysh: No, we want specific version.

bin/base-usage
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh -e
+
+cat <<EOF
+This image serves as the base image for all OpenShift v3 STI builder images.
+It provides all essential libraries and development tools that are needed to
+successfuly build and run the application created using the STI.
+
```
soltysh: It provides all essential libraries and development tools ~~that are~~ needed to successful**l**y build and run ~~the~~ an application ~~created~~ using ~~the~~ STI.

Dockerfile.rhel7
jhadvig: I dont think we want a specific version of rhel7, other then the latest.  We specify the version of centos, cause the their image is called `centos` and the tag is `centos7`

soltysh:
Other than small description enhancements LGTM.

Dockerfile.rhel7
smarterclayton: Why would we not lock to a RHEL base version?  Because we can build and replace it?

----- Original Message -----

Issue #2, opened 2015-02-13
Import RPM keys when building


mnagy:
Run rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 to avoid warning when first running yum install.

mnagy:
Note: No need to address this, I have a fix prepared. Will test it next week and submit PR, this issue is just a reminder for myself.

PR #8, opened 2015-02-19
Import keys through rpmkeys to avoid warning

mnagy:
Fixes issue #2

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,8 @@ MAINTAINER Jakub Hadvig <jhadvig@redhat.com>
 
 # This is the list of basic dependencies that all language Docker image can
 # consume.
-RUN yum install -y --setopt=tsflags=nodocs \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -5,7 +5,8 @@ MAINTAINER Jakub Hadvig <jhadvig@redhat.com>
 
 # This is the list of basic dependencies that all language Docker image can
 # consume.
-RUN yum install -y --setopt=tsflags=nodocs \
+RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
+    yum install -y --setopt=tsflags=nodocs \
```
mfojtik: isn't that key imported automatically by yum when doing yum install?

mnagy: Yeah, but it produces a red warning.

PR #16, opened 2015-03-27
Added scl-utils package

soltysh:
Apparently `scl-utils` is not installed anymore by default both in rhel7 and centos7. 

//cc @mfojtik @jhadvig @mnagy @bparees


mfojtik:
LGTM

jhadvig:
LGTM

PR #20, opened 2015-05-08
automatically squash sti-base

bparees:


bparees:
@soltysh ptal.

bparees:
@goldmann thanks for the awesome squashing tool.

goldmann:
You welcome! If you find any issues or have RFEs - please [file a ticket](https://github.com/goldmann/docker-scripts/issues).

hack/build.sh
```diff
@@ -5,6 +5,23 @@ OS=$1
 
 IMAGE_NAME=openshift/base-${OS}
 
+function squash {	
+	# install the docker layer squashing tool
+	rm -rf docker-scripts
+	git clone https://github.com/goldmann/docker-scripts/
```
goldmann: Please make sure you use a tag. Probably I'll introduce some major changes to the structure of the repo.

hack/build.sh
```diff
@@ -5,6 +5,23 @@ OS=$1
...
+	pip=$(which pip)
+	if [ ! -x $pip ]; then 
+		wget https://bootstrap.pypa.io/get-pip.py
```
soltysh: You could use `easy_install`, like we do in sti-python, see [here](https://github.com/openshift/sti-python/blob/master/3.3/Dockerfile#L30).

soltysh:
Awesome job @bparees, thx @goldmann for the tool :+1: 
LGTM

hack/build.sh
goldmann: BTW, I work on publishing it to PyPi, but this will affect how you call the tool. I'll let you know once it's finished.

mfojtik:
LGTM (@goldmann, how about rewriting the tool to Go? :-)

goldmann:
@mfojtik Haha! No. :)

mfojtik:
@goldmann (sad) :panda_face:

mfojtik:
@goldmann that will allow to distribute it as a single binary without dependencies and it will allow us to use it in STI or other tooling :-)

goldmann:
Having it in Python makes it easy to integrate with rel-eng tools.

soltysh:
@mfojtik with @goldmann's approval we could include it as part of STI in future versions.

mfojtik:
@soltysh if we rewrite this to Go :-)

mfojtik:
@soltysh it will be cool feature of STI to squash the base image and then just add the application layer. So you can have just base and application layers.

goldmann:
@mfojtik Do not touch base layers, this will create a lot of unnecessary data to move around since squashed layers will have new IDs. Squash only what you add (application layer).

mfojtik:
@goldmann in that case squashing in STI does not make sense, because we only add the application layer which is just one layer (I believe).

soltysh:
@mfojtik if so, we should apply the squashing just to our builder images, like @bparees is doing that with the base one.

bparees:
@soltysh we discussed that we don't want to squash our builder images because we don't want minor changes to the assemble/run scripts to force a full re-pull of the entire builder layer.

however now that i restructured the builder layers to put the ADD early in the dockerfile (in order to reduce the number of layers) we may not have that benefit any more anyway.  But we do only introduce 3-4 layers in our builders now.

soltysh:
@mfojtik btw, just found that https://github.com/jwilder/docker-squash

hack/build.sh
bparees: thanks, will do.

bparees:
@soltysh i only looked at it briefly but, that tool doesn't seem to support merging an arbitrary subset of layers like @goldmann 's tool does, which means we couldn't keep centos as its own set of layers.

PR #25, opened 2015-05-18
Add findutils to list of packages

rhcarvalho:
Because "You should install findutils because it includes tools that
are very useful for finding things on your system."
(yum info findutils)

------------
I need `find` in the Python image. @bparees @mfojtik PTAL.


bparees:
will merge once the test build passes.

PR #27, opened 2015-05-26
Added support for io.openshift.sti.scripts-url label

soltysh:
I've added support for `io.openshift.sti-scripts-url` label, but I'm keeping `STI_SCRIPTS_URL` environment variable for compatibility with previous sti and openshift builder versions.

/fyi @bparees @mnagy @jhadvig @mfojtik @kargakis @rhcarvalho

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,9 +4,12 @@ FROM centos:centos7
 MAINTAINER Jakub Hadvig <jhadvig@redhat.com>
 
 # Location of the STI scripts inside the image
+LABEL io.openshift.sti.scripts-url=image:///usr/local/sti
+# Deprecated. Use above LABEL instead, because this will be removed in future versions.
+ENV STI_SCRIPTS_URL=image:///usr/local/sti
+
 # The $HOME is not set by default, but some applications needs this variable
-ENV STI_SCRIPTS_URL=image:///usr/local/sti \
-    HOME=/opt/openshift/src \
+ENV HOME=/opt/openshift/src \
     PATH=/opt/openshift/src/bin:/opt/openshift/bin:/usr/local/sti:$PATH
 
 # This is the list of basic dependencies that all language Docker image can
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -4,14 +4,16 @@ FROM rhel7
 MAINTAINER Jakub Hadvig <jhadvig@redhat.com>
 
 # Location of the STI scripts inside the image
+LABEL io.openshift.sti.scripts-url=image:///usr/local/sti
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -4,9 +4,12 @@ FROM centos:centos7
 MAINTAINER Jakub Hadvig <jhadvig@redhat.com>
 
 # Location of the STI scripts inside the image
+LABEL io.openshift.sti.scripts-url=image:///usr/local/sti
```
mfojtik: is the '=' sign required here? (just my personal preference ;-)

soltysh: Srsly, I have no idea, I've copied that from our images you created :wink: I need to check that, though.

soltysh: So looking at [docker docs](https://docs.docker.com/userguide/labels-custom-metadata/) I see that they are mentioning `=` everywhere, but I just tested and it also works without it. But since we're using equal sign with ENVs, I think we should stick to that standard :smile:

mfojtik: /me hides

i can't help it but that line hurts my eyes :crying_cat_face:

mfojtik: @soltysh we use it in ENV because it is required when you do multiline

soltysh: It's not that bad :sunglasses:

mfojtik: alright :)

jhadvig:
LGTM :goberserk:

Dockerfile.rhel7
```diff
@@ -4,14 +4,17 @@ FROM rhel7
...
 # The $HOME is not set by default, but some applications needs this variable
 # TODO: There is a bug in rhel7.1 image where the PATH variable is not exported
 # properly as Docker image metadata, which causes the $PATH variable do not
 # expand properly.
-ENV STI_SCRIPTS_URL=image:///usr/local/sti \
-    HOME=/opt/openshift/src \
+ENV HOME=/opt/openshift/src \
     PATH=/opt/openshift/src/bin:/opt/openshift/bin:/usr/local/sti:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
-#ENV PATH             /opt/openshift/src/bin:/opt/openshift/bin:/usr/local/sti:$PATH
+#ENV PATH /opt/openshift/src/bin:/opt/openshift/bin:/usr/local/sti:$PATH
```
rhcarvalho: Can't we just remove this commented-out line? The paths included here are also set in the line above.

soltysh: Probably we should, as this setting will be permanent for rhel7 image.

rhcarvalho:
LGTM

PR #33, opened 2015-06-09
Add BASH_ENV so $HOME/.bashrc is sourced automatically

mnagy:
This will allow us to remove all `source .bashrc` lines from the
downstream images.

@mfojtik PTAL. I tested it with sti-perl and tests are passing :)


bparees:
lgtm.

PR #34, opened 2015-06-10
Revert "Add BASH_ENV so $HOME/.bashrc is sourced automatically"

bparees:
This reverts commit c8bc73c3fbea6836505e8e8539be9e0b4fc24c70.


bparees:
@mnagy @mfojtik this change caused some recursive behavior in the ruby image.  can you investigate today?  for now i'm backing it out so we can get working builds again. 

@mfojtik this the fix for now for the issue i emailed you about.  no need to talk to the SCL guys i guess.

PR #35, opened 2015-06-10
Enable SCL globally without sourcing bashrc

mfojtik:
This PR enables the SCL collections globally by using BASH_ENV to inject the `scl enable` to every `/bin/bash` session.

This require change to every image we have, so don't merge it on its own, this have to be merged together with other PR's I will link to this PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ ENV STI_SCRIPTS_URL=image:///usr/local/sti
 
 # The $HOME is not set by default, but some applications needs this variable
 ENV HOME=/opt/openshift/src \
+    BASH_ENV=/opt/openshift/src/.bashrc \
     PATH=/opt/openshift/src/bin:/opt/openshift/bin:/usr/local/sti:$PATH
 
 # This is the list of basic dependencies that all language Docker image can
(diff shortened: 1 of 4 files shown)


mfojtik:
FYI: The tests will fail for all those images without the change in sti-base. I'm testing all PR's manually using make test.

mfojtik:
@bparees @mnagy @rhcarvalho @soltysh @jhadvig guys, can I ask you to just double-check if I didn't miss something in all this copy & pasting? I ran `make test` for every repo we have with success.

Also I just want to make you aware of this change. This means that we no longer need to source `.bashrc` in any script and the SCL should be enabled automatically in every bash session.

@hhorak can you please confirm that using BASH_ENV this way won't break anything? So far all our tests are passing (including the full examples we have in openshift org).

mnagy:
LGTM

rhcarvalho:
LGTM, love the SKIP_SQUASH.

mfojtik:
I should probably put a fat warning to every .bashrc saying that you should NOT add more stuff there as it will be run everytime you execute the bash itself. This can be nasty and for example in sti-python we do the arbitrary user magic in that file. This magic will be execute everytime you run bash. We should probably move that into separate file and source it just in `run` script as I think that it is the only place where we do care about setting the arbitrary user.

mfojtik:
(warning added), got 2 lgtm, merging this.

soltysh:
Nicely done @mfojtik !

PR #40, opened 2015-06-10
Move SCL enable to /opt/openshift/etc/scl_enable

mfojtik:
Details: https://github.com/openshift/sti-base/issues/38

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ ENV HOME=/opt/openshift/src \
 # When bash is started non-interactively, to run a shell script, for example it
 # looks for this variable and source the content of this file. This will enable
 # the SCL for all scripts without need to do 'scl enable'.
+ADD contrib/scl_enable /opt/openshift/etc/scl_enable
 ENV BASH_ENV=/opt/openshift/etc/scl_enable \
     ENV=/opt/openshift/etc/scl_enable \
     PROMPT_COMMAND=". /opt/openshift/etc/scl_enable"
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,8 +10,14 @@ ENV STI_SCRIPTS_URL=image:///usr/local/sti
 
 # The $HOME is not set by default, but some applications needs this variable
 ENV HOME=/opt/openshift/src \
-    BASH_ENV=/opt/openshift/src/.bashrc \
     PATH=/opt/openshift/src/bin:/opt/openshift/bin:/usr/local/sti:$PATH
+
+# When bash is started non-interactively, to run a shell script, for example it
+# looks for this variable and source the content of this file. This will enable
+# the SCL for all scripts without need to do 'scl enable'.
+ENV BASH_ENV=/opt/openshift/etc/scl_enable \
+    ENV=/opt/openshift/etc/scl_enable \
+    PROMPT_COMMAND=". /opt/openshift/etc/scl_enable"
 
 # This is the list of basic dependencies that all language Docker image can
 # consume.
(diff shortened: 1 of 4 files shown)


bparees:
@mfojtik fyi i'm messing w/ the sti-base PR tester so your test build may fail... push another change or rerun the failed job if you want a good test.

mfojtik:
@bparees it will fail anyway... it requires other PR merged ;-)

Dockerfile
```diff
@@ -10,9 +10,15 @@ ENV STI_SCRIPTS_URL=image:///usr/local/sti
 
 # The $HOME is not set by default, but some applications needs this variable
 ENV HOME=/opt/openshift/src \
-    BASH_ENV=/opt/openshift/src/.bashrc \
     PATH=/opt/openshift/src/bin:/opt/openshift/bin:/usr/local/sti:$PATH
 
+# When bash is started non-interactively, to run a shell script, for example it
+# looks for this variable and source the content of this file. This will enable
+# the SCL for all scripts without need to do 'scl enable'.
+ENV BASH_ENV=/opt/openshift/etc/scl_enable \
```
mfojtik: just to explain this:

`BASH_ENV` is used for non-interactive shell (/bin/bash -c foo)
`ENV` is used for non-interactive restricted shell (/bin/sh -c foo)
`PROMPT_COMMAND` is execute in interactive shell before displaying PS1 (/bin/bash -i, or just /bin/bash)

mfojtik: @mnagy review this pls.

bparees:
ok i get it now.  thanks for the explanation @mfojtik.

i'm good with this pending adding the various scl enablement tests to our individual images.

Dockerfile
```diff
@@ -10,9 +10,15 @@ ENV STI_SCRIPTS_URL=image:///usr/local/sti
...
+    ENV=/opt/openshift/etc/scl_enable \
+    PROMPT_COMMAND=". /opt/openshift/etc/scl_enable"
```
mnagy: This will be executed for every prompt, which seems excessive to me. Also, why is it needed when BASH_ENV and ENV should take care of that?

Btw, since you're getting so serious about this, why not also handle the bash -l case? When bash is invoked as a login shell, it'll read /etc/profile, which in turn reads /etc/profile.d/*.sh, so all you need to do is copy the scl_enable to /etc/profile.d with added .sh suffix.

mfojtik: I'm going to unset the PROMPT_COMMAND in scl_enable which should be more safe.

The login shell is covered by PROMPT_COMMAND (tested).

rhcarvalho:
@mfojtik if we set `BASH_ENV`, `ENV`, and `PROMPT_COMMAND` in sti-base, we create a sort of inverted dependency, so that all images based on sti-base need to 1) provide that file, 2) unset those variables.
Discussing with @mnagy we thought of two solutions:
- Provide a file `/opt/openshift/etc/scl_enable` in sti-base that would just unset those variables; OR
- Do not set these variables in sti-base, but do so for the sti-* images we provide.

What do you think?

mfojtik:
@rhcarvalho I will prefer 1) for now and 2) in long term.

bparees:
@rhcarvalho @mfojtik if the file is not present does it actually fail?

mfojtik:
@bparees you will get errors when you run just the sti-base in interactive bash (PROMPT_COMMAND). It will work but the feng-shui will be ruined...

bparees:
@mfojtik got it. the dummy file that can be overwritten by the downstream images works for me.

mfojtik:
@bparees yes... it means you can still base your image on sti-base and we don't break you...

bparees:
"5 stars, would merge again!"

PR #41, opened 2015-06-11
Added mandatory labels

eliskasl:
@brenton, could you have a quick look please? This was more or less a a wild guess.

### DIFF
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -8,6 +8,12 @@ LABEL io.s2i.scripts-url=image:///usr/local/sti
 # Deprecated. Use above LABEL instead, because this will be removed in future versions.
 ENV STI_SCRIPTS_URL=image:///usr/local/sti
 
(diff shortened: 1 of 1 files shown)


bparees:
v1.0 seems ok.  probably important to note that this image won't actually be available to customers I don't think.  it's just for us to build on top of.  (I suppose in theory we might want to let customers build on top of it at some point though)

Dockerfile.rhel7
```diff
@@ -8,6 +8,12 @@ LABEL io.s2i.scripts-url=image:///usr/local/sti
 # Deprecated. Use above LABEL instead, because this will be removed in future versions.
 ENV STI_SCRIPTS_URL=image:///usr/local/sti
 
+# Labels consumed by Red Hat build service
+LABEL Component="rhel-server-docker" \
```
brenton: For now I think the Component should be `openshift-sti-base-docker` since that matched the dist-git repo.

PR #47, opened 2015-06-20
do not skip_squash on tests by default

bparees:


bparees:
@mfojtik @soltysh @jhadvig @mnagy @rhcarvalho this is on by default now because we broke the squash logic which wasn't caught until the push job ran.  if you don't want to do squashing, set a SKIP_SQUASH=1 env variable in your environment.

i'm updating the vagrant test logic to directly invoke make test so we'll catch this stuff in the future before it's merged.

PR #49, opened 2015-06-23
Replacing centos7 base with origin-base

jhadvig:
Base on https://github.com/openshift/sti-base/issues/30

I think I would be also appropriate to change the base-rhel7 base image accordingly, since this change introduces inconsistency between base-centos7 and base-rhel7, but since we dont have any openshift/origin-base-rhel7 we will need to keep the change in mind.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM centos:centos7
+FROM openshift/origin-base
 
 # This image is the base image for all OpenShift v3 language Docker images.
 MAINTAINER Jakub Hadvig <jhadvig@redhat.com>
@@ -30,12 +30,10 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
   autoconf \
   automake \
   bsdtar \
-  epel-release \
   findutils \
   gcc-c++ \
   gdb \
   gettext \
-  git \
   libcurl-devel \
   libxml2-devel \
   libxslt-devel \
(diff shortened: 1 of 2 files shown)


jhadvig:
@bparees @smarterclayton @bparees thoughts ?

smarterclayton:
We should create a base rhel dockerfile in origin

jhadvig:
@smarterclayton https://github.com/openshift/origin/pull/3397 fyi

bparees:
let's hold this until the base-rhel7 image exists and have this PR take advantage of that too.

jhadvig:
@bparees +1
I've updated the origin PR which creates the rhel7 based image. So the images naming will be following:
- centos7 -> openshift/origin-base
- rhel7     -> openshift/ose-base

When that PR will land, I will update also the sti-base Dockerfile.rhel7

bparees:
looks like that PR landed.  i'm rerunning the test on this one.

jhadvig:
@bparees I've updated the Dockerfile.rhel7 so its based on openshift/ose-base

bparees:
@jhadvig merge build failed..I didn't look into why but i'm guessing the openshift/ose image isn't available on the instance and you need to update vagrant to explicitly pull it?

bparees:
@jhadvig or get the ami builder to include it.

mfojtik:
@bparees @jhadvig I assume that you guys are also aware that platform will have to make openshift/ose and origin-base built in OSBS (the same for sti-base). @eliskasl @hhorak FYI.

bparees:
@mfojtik yes, aware but probably not taking into account :)

bparees:
@jhadvig @mfojtik given that the SCL team is going to create/own their own base image going forward, i think it's probably safe to abandon this (sorry).  what do you think?

mfojtik:
+1

PR #52, opened 2015-06-29
Change owner of /opt/openshift in base image

mfojtik:
This will allow to drop the `chown` from all Dockerfiles (as long as you copy the files in the directory **after** you drop to USER 1001. 

This can be merged without any harm (as the existing images will just do chown anyway. I'll be removing the chown in the other PR (namespace fix). So this should go in before that...


soltysh:
LGTM

PR #53, opened 2015-06-29
Fix comparison in hack/build.sh

mfojtik:
Fixes these errors:

```console
...
Successfully built c4c003e04662
hack/build.sh: line 22: [: : integer expression expected
Building and testing openshift/base-rhel7:e651aac ...
...
```

(iow. the case shen SKIP_SQUASH is not defined)


mfojtik:
@bparees FYI (will fix this in all build scripts... dunno how this sneaked out from `set -e`... I guess because the ./hack/build.sh does not `set -e` and it doesn't inherit it from parent bash process.

PR #58, opened 2015-07-16
Rename /opt/openshift to opt/app-root 

humaton:
https://github.com/sclorg/rhscl2dockerfile/issues/29

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,16 +14,16 @@ LABEL io.s2i.scripts-url=image:///usr/local/sti
 ENV STI_SCRIPTS_URL=image:///usr/local/sti
 
 # The $HOME is not set by default, but some applications needs this variable
-ENV HOME=/opt/openshift/src \
-    PATH=/opt/openshift/src/bin:/opt/openshift/bin:/usr/local/sti:$PATH
+ENV HOME=/opt/app-root/src \
+    PATH=/opt/app-root/src/bin:/opt/app-root/bin:/usr/local/sti:$PATH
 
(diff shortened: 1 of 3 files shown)


jhadvig:
@humaton can you also update the path in the `base-usage` file.
https://github.com/openshift/sti-base/blob/master/bin/base-usage#L19

mfojtik:
[test]

humaton:
Sorry guys but I would not write tests in bash.

mfojtik:
@humaton don't worry, I'm just using this PR as canary to test Jenkins ;-)

mfojtik:
[test]

mfojtik:
[test]

jhadvig:
@humaton pls squash it into one commit, no point to have two separate commits with the same change.

jhadvig:
LGTM

mfojtik:
I would want to test all images before we merge this change so we know if we're broken or not.

jhadvig:
@mfojtik +1
will manually test this change in all our images

jhadvig:
@bparees @mfojtik I've applied and tested(manually) the change on all our sti images and it worked. Since that I've opened PR for sti-ruby,python,perl,php
Will also test the change on *-ex apps so we are safe on this change, since this is a pretty big change.

bparees:
:+1: thanks @jhadvig -ex testing is a very good idea

jhadvig:
Tested this change with all our sti images + rails/dancer/django-ex
Everything went without problem. Lets merge it :)

PR #59, opened 2015-07-21
Update with latest OSBS requirements

sdodson:


bparees:
lgtm, will merge after test build passes.

PR #60, opened 2015-07-27
Use /usr/bin for sti scripts

hhorak:
Since we're trying to come up with guidelines and best practices for creating docker images, I'm also thinking about things like locations of files within the container. This is a proposal to change location of sti scripts.

The reasons for this change are:
* what FHS says about '/usr/local':
The '/usr/local' hierarchy is for use by the system administrator when installing software locally. It needs to be safe from being overwritten when the system software is updated. It may be used for programs and data that are shareable amongst a group of hosts, but not found in '/usr'. (http://www.pathname.com/fhs/pub/fhs-2.3.html#USRLOCALLOCALHIERARCHY)
* we may include the sti scripts in the RPMs once (in that case it would probably be links managed by alternatives, since various implementations of /usr/bin/run will be different across images)
* we need to have those commands available for and run by users, so I don't see any special reason to handle them differently than other commands

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,17 +5,17 @@ MAINTAINER Jakub Hadvig <jhadvig@redhat.com>
 
 # Location of the STI scripts inside the image
 #
-LABEL io.openshift.s2i.scripts-url=image:///usr/local/sti
+LABEL io.openshift.s2i.scripts-url=image:///usr/bin
 
 # DEPRECATED: This label will be kept here for backward compatibility
-LABEL io.s2i.scripts-url=image:///usr/local/sti
+LABEL io.s2i.scripts-url=image:///usr/bin
 
 # Deprecated. Use above LABEL instead, because this will be removed in future versions.
-ENV STI_SCRIPTS_URL=image:///usr/local/sti
+ENV STI_SCRIPTS_URL=image:///usr/bin
 
 # The $HOME is not set by default, but some applications needs this variable
 ENV HOME=/opt/app-root/src \
-    PATH=/opt/app-root/src/bin:/opt/app-root/bin:/usr/local/sti:$PATH
+    PATH=/opt/app-root/src/bin:/opt/app-root/bin:$PATH
 
 # When bash is started non-interactively, to run a shell script, for example it
 # looks for this variable and source the content of this file. This will enable
@@ -69,7 +69,7 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
 
 # Create directory where the image STI scripts will be located
 # Install the base-usage script with base image usage informations
-ADD bin/base-usage /usr/local/sti/base-usage
+ADD bin/base-usage /usr/bin/base-usage
 
 # Directory with the sources is set as the working directory so all STI scripts
 # can execute relative to this path
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ LABEL io.s2i.scripts-url=image:///usr/libexec/s2i
 # Deprecated. Use above LABEL instead, because this will be removed in future versions.
 ENV STI_SCRIPTS_URL=image:///usr/libexec/s2i
 
-# Path to be used in other layers to place s2i scritps into
+# Path to be used in other layers to place s2i scripts into
 ENV STI_SCRIPTS_PATH=/usr/libexec/s2i
 
 # The $HOME is not set by default, but some applications needs this variable
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,13 +5,16 @@ MAINTAINER Jakub Hadvig <jhadvig@redhat.com>
 
 # Location of the STI scripts inside the image
 #
-LABEL io.openshift.s2i.scripts-url=image:///usr/bin
+LABEL io.openshift.s2i.scripts-url=image:///usr/libexec/s2i
 
 # DEPRECATED: This label will be kept here for backward compatibility
-LABEL io.s2i.scripts-url=image:///usr/bin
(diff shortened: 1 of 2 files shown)


bparees:
this is going to require downstream changes in the s2i image assemble scripts.  i'd like to at least hold off on this until we get 3.0.1 out the door.

bparees:
specifically these lines:
https://github.com/openshift/sti-python/blob/master/3.3/Dockerfile#L19

we used to use ONBUILD to avoid this problem but that introduced its own complications.

hhorak:
There is already `STI_SCRIPTS_URL` variable defined. If we have also `STI_SCRIPTS_PATH`, we could change it once to `COPY .sti/bin $STI_SCRIPTS_PATH` and avoid similar issues in case the path changes again..

bparees:
the STI_SCRIPTS_URL env variable was deprecated in favor of labels.  Which we recognized had the negative impact of making it impossible to use from assemble scripts.

but yes we can introduce a scripts_path env variable in the base image that's not part of the s2i API, but is used by our downstream images' Dockerfiles.

mfojtik:
@hhorak @bparees i'm a little bit worried that having `/usr/bin/run` and `/usr/bin/assemble` can lead into conflict with other RPM's. Also moving the scripts to that location makes them automatically available in PATH (not sure if it can break something, but I have mixed feelings about that ;-)

I'm fine with following FHS and 'standardizing' the way how files are organized in our Docker images, but I also think that we should be careful.

bparees:
agree, particularly given how generic "run" is.

@hhorak is there another option here?

hhorak:
If the potential conflict is the issue, we may probably use /usr/lib/sti/bin, similar to what qt uses. And add /usr/lib/sti/bin into path in dockerfiles...

bparees:
yeah i think that would be better. (to avoid the conflict)

mfojtik:
+1 for `/usr/lib/sti/bin` or `/usr/libexec/sti/bin` (we have at least more control on what will be in PATH)

bparees:
@hhorak can you update this to use /usr/lib/sti/bin and introduce an env var that the downstream images can use for the COPY command?

bparees:
@hhorak actually let's make it /usr/lib/s2i/bin  @mfojtik agree?

mfojtik:
@bparees to me /usr/lib is strange because it usually doesn't contain any binaries, just .so files. The /usr/libexec is more appropriate (that is for example where gcc/go stores binaries).

hhorak:
So, `/usr/libexec/s2i` and having this path in `$PATH`? It seems less correct to me:
`/usr/libexec` includes internal binaries that are not intended to be executed directly by users or shell scripts. (http://www.linuxbase.org/betaspecs/fhs/fhs/ch04s07.html)

hhorak:
I think there are not more than two options here:
(1) either run/usage/build are expected to be run directly by users within the container; in that case those libraries cannot conflict with any other package (they don't now and I don't believe any other package will start using such general binary names)
(2) or run/usage/build are only expected to be run from outside using `source-to-image` (sti) tool; in that case they don't have to be in `$PATH` and placing them into `/usr/libexec/s2i` is totally correct

Which of the two options above do we talk here? AFAICT it is about (2).

But if we want to allow to do something like:
```
#> docker run -ti ... bash
docker> run
```
... is there a possibility to rename those libraries to something like `s2i-run`, `s2irun` insider the container or whatever like that? Then sti tool would simply use `$STI_SCRIPTS_URL/run` and users would call `s2i-run` when running the commands directly (without source-to-image).

bparees:
@hhorak i'm really not sure why we are adding them to the path today.  @mfojtik @soltysh can you think of a reason we did that?  we should be invoking them with an absolute path anyway when s2i runs them.

and no we wouldn't expect a user to invoke run/assemble/save-artifacts/usage directly.  so yeah, I think we're in scenario (2)

soltysh:
@bparees I don't remember why we ended up with adding those scripts to `$PATH` but I'm positive ([see here](https://github.com/openshift/source-to-image/blob/master/pkg/docker/docker.go#L370-L376)) we are prefixing the script name with a full path to it, so I'm ok with @hhorak scenario (2).

bparees:
@hhorak so let's put them in /usr/libexec/s2i and set an env variable pointing there.

the downstream images can then copy their scripts to that location using the env variable.

can you update this PR to do that and i'll work on PRs for the downstream images and then we'll have to merge it all at once.

Dockerfile
```diff
@@ -5,17 +5,20 @@ MAINTAINER Jakub Hadvig <jhadvig@redhat.com>
...
 
 # DEPRECATED: This label will be kept here for backward compatibility
-LABEL io.s2i.scripts-url=image:///usr/local/sti
+LABEL io.s2i.scripts-url=image:///usr/libexec/s2i
 
 # Deprecated. Use above LABEL instead, because this will be removed in future versions.
-ENV STI_SCRIPTS_URL=image:///usr/local/sti
+ENV STI_SCRIPTS_URL=image:///usr/libexec/s2i
+
+# Path to be used in other layers to place s2i scritps into
```
bparees: scripts

hhorak: fixed

bparees:
s2i image update PRs:
https://github.com/openshift/sti-perl/pull/51
https://github.com/openshift/sti-nodejs/pull/69
https://github.com/openshift/sti-php/pull/59
https://github.com/openshift/sti-python/pull/67
https://github.com/openshift/sti-ruby/pull/76
https://github.com/openshift/sti-wildfly/pull/59

@mfojtik please sanity check.

jhadvig:
LGTM

PR #61, opened 2015-08-03
Update the structure of Makefile and hack/build.sh to match our other image repos

jhadvig:
Updated the structure of `Makefile` and `hack/build.sh` to match our other image repos.
@bparees @mfojtik PTAL


mfojtik:
LGTM

PR #64, opened 2015-08-31
Add entrypoint so the commands as CMD on non-standard path are found

hhorak:
With current sti-base image the following runs python 2.7, while one would expect python 3.3:

  #> docker pull openshift3/python-34-rhel7
  #> docker run --rm -ti openshift3/python-34-rhel7 python --version
  Python 2.7.5

Adding entrypoint will ensure correct PATH is set also at the time the CMD is being found.


mfojtik:
[test]

bparees:
[test]

bparees:
lgtm.

PR #65, opened 2015-09-05
default to group 0

bparees:
this will ensure source code is writable by the uid running the container on openshift (the container uid is always a member of gid 0).  This will allow users to docker exec  (or oc exec) into the running container and manipulate source code to do dynamic debugging/testing.

previously the source code was only writable by uid 1001/gid 1001 which never corresponded to the uid/gid running the container once it was deployed in openshift.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,10 +62,9 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
   zlib-devel && \
   yum clean all -y && \
   mkdir -p ${HOME} && \
-  groupadd -r default -f -g 1001 && \
-  useradd -u 1001 -r -g default -d ${HOME} -s /sbin/nologin \
+  useradd -u 1001 -r -g 0 -d ${HOME} -s /sbin/nologin \
       -c "Default Application User" default && \
-  chown -R 1001:1001 /opt/app-root
+  chown -R 1001:0 /opt/app-root
 
 # Create directory where the image STI scripts will be located
 # Install the base-usage script with base image usage informations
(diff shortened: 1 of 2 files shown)


bparees:
[test]

bparees:
https://github.com/openshift/sti-wildfly/pull/58
https://github.com/openshift/sti-php/pull/56
https://github.com/openshift/sti-ruby/pull/70
https://github.com/openshift/sti-nodejs/pull/66
https://github.com/openshift/sti-perl/pull/47/files
https://github.com/openshift/sti-python/pull/63

bparees:
these will all need to be merged together, tests won't pass if they are merged independently, but the python PR isn't quite ready yet.

bparees:
ok i think python is fixed too now.  still want to test this with our ex repos though.

hhorak:
@bparees Please, provide proper reasoning why this change is done.

soltysh:
Was about to post exactly the same question, why you're doing it?

jhadvig:
@soltysh check discussion here: https://github.com/openshift/sti-perl/pull/47

bparees:
@hhorak @soltysh description updated.  and @jhadvig thanks.

bparees:
@mfojtik @jhadvig ok all the PRs should be good to go, please take a look.  again, tests won't pass until they all merge at once.

i've tested w/ our example repos as well.

mfojtik:
@bparees OK, but you will have to explain me why we are doing this in person :-)

LGTM

jhadvig:
@bparees seems like everybody whats your explanations. You should start using twitter for this :smile:  

lgtm, but will wait with merging for you :)

PR #66, opened 2015-09-08
Enable rhel-7-server-optional-rpms for bsdtar

sdodson:
I don't know if this is necessary for centos or not

### DIFF
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -39,7 +39,8 @@ ENV BASH_ENV=/opt/app-root/etc/scl_enable \
 # Also setup the 'openshift' user that is used for the build execution and for the
 # application runtime execution.
 # TODO: Use better UID and GID values
-RUN yum install -y --setopt=tsflags=nodocs \
+RUN yum-config-manager --enable rhel-7-server-optional-rpms && \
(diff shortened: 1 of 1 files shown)


mfojtik:
[test]

mfojtik:
@sdodson we will see

bparees:
what does bsdtar provide that tar isn't providing?

mfojtik:
@bparees it makes ruby bundler happy! (or at least @jhadvig said so)

bparees:
failure: Error getting repository data for rhel-7-server-optional-rpms, repository not found

bparees:
"yum-config-manager --enable rhel-7-server-optional-rpms" is how we do this in our other images:
https://github.com/openshift/mongodb/blob/master/2.4/Dockerfile.rhel7#L36

sdodson:
Switched to using yum-config-manager. Though I doubt that'll fix the test.

bparees:
i'm confused.:
https://github.com/openshift/sti-base/blob/master/Dockerfile.rhel7#L74

https://github.com/openshift/sti-base/pull/66/files#diff-b0556bed298b3a4fb269eb10d69c91a5L75

i don't see why it should matter, but do a rebase?

PR #70, opened 2015-10-23
Add fix-permissions script

csrwng:
Adds a script to fix source directory permissions to be used in all derived images.
Also change the ADD instruction to COPY in Dockerfiles


csrwng:
@bparees @jhadvig

bparees:
lgtm but will await test results.

bparees:
@kconner @rcernich we found that in addition to setting the ownership of the files to the root group and making it group RW, it's also important to make the directories group executable.  You will want to do this in the middleware images also, in order to have a good experience with oc sync.

we're going to invoke this new script from our assemble scripts after everything has been built/copied.

PR #71, opened 2015-11-04
sti -> s2i change

hhorak:
Base usage script does still include sti instead of s2i.


bparees:
nice catch @hhorak.  lgtm.

PR #73, opened 2015-11-30
Rename MAINTAINER

jhadvig:
As all other images have the same `MAINTAINER`, so should this. Right ?
@bparees ptal :)


bparees:
The Scl team is likely going to create their own base image in the near future, not sure I'd consider them the maintainers of this one. If you just want your name off it since you switched teams, you can put my name on it.

jhadvig:
Just thought that all the our images should have the same MAINTAINER. Going to close this. Thanks

PR #77, opened 2016-02-17
Add LC_ALL and TERM env vars

caruccio:


bparees:
@caruccio Can you add this to the Dockerfile.rhel7 also?

@mfojtik @rhcarvalho can you think of any reason setting these env vars would be a bad idea?

caruccio:
Oops, my bad. Added and amended.

mfojtik:
@bparees @caruccio I'm OK with LC_ALL but what is the reason for setting TERM to vt100 ? Does that mean if I `oc rsh` or `docker exec` into container and I use different terminal, the console will be broken?

caruccio:
@mfojtik In fact you already need to set TERM to run most of the apps, like top, watch, vi.
vt100 is just a pretty standard terminal with small capabilities. If your terminal has more capabilities, just set it (or define it into .sti/environment). You should do it anyway.

rhcarvalho:
## the values today

Running bash on a sample ruby app image:

```console
bash-4.2$ printenv TERM  
xterm                    
bash-4.2$ printenv LC_ALL
```

## the change

@bparees I think both env vars are too user & application specific to be part of `sti-base`.

Both can change the behavior of applications, and it is hard to tell what should be the default value enforced to all S2I users.

Why `LC_ALL=en_US.UTF-8` and not `LC_ALL=C` or any other value? I would argue that if your application needs a certain value for `LC_ALL`, write that in `.sti/environment` and you should be good to go.

Similar question to why making `vt100` the default... the default today is `xterm`, what's wrong with it?

rhcarvalho:
@caruccio can you tell us more about why you need those variables set to the values you are proposing? Perhaps there's some underlying problem we're not talking about and this change would address?

bparees:
@rhcarvalho TERM is being set because when you oc rsh into a pod, the term value is not set and nothing works well.  but now i'm wondering if that's an issue with oc rsh instead of with our images.

mfojtik:
@bparees I think `oc rsh` should pass the TERM to match your terminal.

caruccio:
@rhcarvalho there is nothing special with those values. I only believe `vt100` is a more common (i.e. basic) terminal than `xterm`. The same applies to `LC_ALL=en_US.UTF-8`: I really don't see much applications using `C` as an usable locale. Also, `UTF-8` is becoming the __de facto__ file encoding.

mfojtik:
@caruccio @rhcarvalho UTF-8 is a good default to me. You can always change that to what you want if you set the env in deployment or just pass -E to docker run. IF having 'C' can cause troubles to apps, I would rather go with 'UTF-8'.

caruccio:
These are the values from a plain sti-php pod:

```
$ oc rsh php-1-4crdc
bash-4.2$ printenv TERM
bash-4.2$ printenv LC_ALL
bash-4.2$
```

IMHO, it's a more userfriendly approach to define reasonable defaults.

rhcarvalho:
Hmm.. so `oc rsh` and `docker run` are leading to different values of TERM....

> the term value is not set and nothing works well

@bparees could you elaborate? What do you mean by "nothing works well"? Smells like a problem in `oc  rsh`, since `docker run s2i-powered-image` sets TERM to `xterm`.

If that's the case we should be fixing `oc rsh` and not `sti-base`.

caruccio:
Is it possible for `oc rsh` to set the same `TERM` as the terminal where it is running already?
Also, web console do not sets a terminal. User needs to set it by hand.

rhcarvalho:
Apparently the problem with TERM is this https://github.com/docker/docker/issues/9299.

I would propose to do what @caruccio is saying above, make `oc rsh` set TERM to whatever value it has in the environment where `oc` is run.

Inheriting TERM from the context where `oc` is run is better than baking some value in S2I images, as it will apply to `oc rsh` independent of what image you are running, but dependent on your actual terminal.

mfojtik:
@rhcarvalho we will have to augment the web console terminal to do the same.

rhcarvalho:
@mfojtik yes.

So we have here two issues for https://github.com/openshift/origin:

1. Make `oc rsh` set TERM, circumventing the Docker issue linked above.
  Optional: do the same for `oc exec -t`.
2. Make the Web Console terminal set TERM to whatever value it is reasonable there.

Now we can discuss what's the deal with LC_ALL. @caruccio, could you please explain why do we need all S2I images to have this value set, what's the use case? Thanks!

caruccio:
Ruby 1.9 is not UTF-8 aware. The error below happens because there is a non-ascii character __in a comment__ from a .rb file. It fails, even if the file is explicitly said to be UTF-8 by starting it with `# encoding: utf8`.

```
---> Cleaning up unused ruby gems ...
249	E0218 17:07:20.823823       1 util.go:91] rake aborted!
250	E0218 17:07:20.823876       1 util.go:91] Encoding::UndefinedConversionError: U+00E7 from UTF-8 to US-ASCII
251	E0218 17:07:20.823913       1 util.go:91] 
252	E0218 17:07:20.823922       1 util.go:91] (See full trace by running task with --trace)
253
```

rhcarvalho:
Disclaimer: I am not a Ruby expert. But tried to do some basic research on the problem...

We have official images for Ruby 2.0 and 2.2, so I imagine you're layering a Ruby 1.9 image on top of sti-base?
If this is only affecting Ruby, or even just some particular version of Ruby, I'd make changes just to the lowest layer possible, not all the way up to the parent layer if it is not necessarily applicable to all descendants.

> Ruby 1.9 is not UTF-8 aware.

Again, I'm no Ruby expert, but I know that Yehuda Katz is. According to [this post](http://yehudakatz.com/2010/05/05/ruby-1-9-encodings-a-primer-and-the-solution-for-rails/):
- in Ruby 1.8, Strings are simple sequences of bytes
- Ruby 1.9 announced support for encodings

Now knowing what's special about Ruby 1.9, let me try to reproduce the problem:

First on my host with Ruby 2.2:

```console
$ ruby --version
ruby 2.2.4p230 (2015-12-16 revision 53155) [x86_64-linux]
$ cat foo.rb 
# encoding: utf8

# meu comentário, 好吧？

42
$ file foo.rb 
foo.rb: UTF-8 Unicode text
$ ruby foo.rb 
foo.rb:1: unknown encoding name: utf8 (ArgumentError)
$ printenv LC_ALL
$ vim foo.rb                                                                                           
$ cat foo.rb                                                                                           
# encoding: utf-8

# meu comentário, 好吧？

puts 42
$ ruby foo.rb 
42
```

So the problem so far is that **utf8** is not recognized, but **utf-8** is.

Now let me try that on Ruby 1.9:

```console
$ docker run --rm -it --entrypoint=/bin/bash --workdir=/opt/ruby/src -v=`pwd`:/opt/ruby/src ncdc/ruby-1.9.x-centos
bash-4.1$ scl enable ruby193 nodejs010 bash
bash-4.1$ cat foo.rb 
# encoding: utf-8

# meu comentário, 好吧？

puts 42
bash-4.1$ ruby foo.rb 
42
bash-4.1$ cat bar.rb 
# encoding: utf8

# meu comentário, 好吧？

puts 42
bash-4.1$ ruby bar.rb 
bar.rb:1: unknown encoding name: utf8 (ArgumentError)
bash-4.1$ printenv LC_ALL
```

Once again, all I could get to is that **utf8** is not recognized, but **utf-8** is.


@caruccio maybe it was a typo in your comment? If you would be able to, could you provide a reproducible example?

------------------------

My thoughts so far:

Coming from the Python world, I've seen similar temptations to change the system-wide locale settings to "fix" encoding problems, when in fact they are programming problems...

I am opposed to set LC_ALL in `sti-base` or `sti-ruby` if only for this reason. From what I've read it is nowhere a Ruby requirement to have that set to anything, nor to anything specific.

It can be set per-app if anyone wants to go down that path:
http://stackoverflow.com/questions/5908774/set-global-default-encoding-for-ruby-1-9

In OpenShift, this could be an environment variable in the Build and Deployment configs.

That said, I don't want to be radical, and my opinion can change given further evidence I might be missing.

caruccio:
@rhcarvalho Unfortunately I'm not allowed to expose the cliente code. I'm going to try to isolate the problem here. It's being a busy day here, so can't promise it for today.
Thanks a lot for your help.

rhcarvalho:
@caruccio np. Let me know if you need help.

bparees:
@caruccio bump, any update?

caruccio:
@bparees not yet. I'm gonna try to reproduce and isolate it tomorrow.

caruccio:
@rhcarvalho I've managed to isolate the problem. It's inside a rake task with UTF chars, even if encoding is declared at the very first line. Please note I'm inside a ruby 1.9.3 POD, built from this sti: https://github.com/getupcloud/sti-ruby-extra/tree/master/1.9. The image can be found in https://hub.docker.com/r/getupcloud/ruby-extra-19-centos7/

```
bash-4.2$ cat lib/tasks/encoding_error.rake
# encoding: utf-8
namespace :encoding_error do
  desc "Descrição inválida"
  task :my_task => :environment do
    puts "isto deve falhar"
  end
end
```

Rake encoding is also being set into Rakefile: 
```
bash-4.2$ head Rakefile -n 1
# encoding: utf-8
```
The failling command is https://github.com/getupcloud/sti-ruby-extra/blob/526f1b2fe6990e6beda827b094c5550dc711c856/1.9/s2i/bin/assemble#L9, copied from https://github.com/openshift/sti-ruby/blob/8847438ba06307f86ac877465eadc835201241df/2.0/s2i/bin/assemble#L9

This is my setup:
```
bash-4.2$ bundle --version
Bundler version 1.3.5
bash-4.2$ bundle exec 'rake --version'
rake, version 10.5.0
bash-4.2$ bundle exec 'ruby --version'
ruby 1.9.3p484 (2013-11-22) [x86_64-linux]
```


Listing tasks with UTF chars will fail:
```
bash-4.2$ bundle exec 'rake -T encoding_error:my_task'
rake aborted!
Encoding::UndefinedConversionError: U+00E7 from UTF-8 to US-ASCII

(See full trace by running task with --trace)
```

However when setting LC_ALL, ll goes fine:
```
bash-4.2$ LC_ALL=en_US.UTF-8 bundle exec 'rake -T encoding_error:my_task'
rake encoding_error:my_task  # Descrição inválida
```

rhcarvalho:
@caruccio thanks for the detailed comment, and for working on a Ruby 1.9 image :+1: 

I'll try to reproduce it here.

In any case, initial research on the Internet suggests this might affect only 1.9, and not 2.x. Have you tried it? I'll try myself.

I still think changing LC_ALL doesn't belong in sti-base, but in a lower layer, either builder image or specific project image -- "en_us.UTF-8" possibly doesn't satisfy everybody.
What do you think?

caruccio:
@rhcarvalho Yes, I can confirm it affects only 1.9.

IMHO, a sensible default should be set by sti-base, just like one expect to have something set in a full OS install. When LC_ALL is explicitly set, users can easily spot any issue related to locale/encoding. For example, it took me a while until I realize LC_ALL was the missing piece for my specific problem.
Even if `en_US.UTF-8` is not the perfect choice, it's still a reasonable value for most projects (AFAIK).

rhcarvalho:
> a sensible default should be set by sti-base, just like one expect to have something set in a full OS install.

AFAICT LC_ALL is not set by default by Linux distros (nor Docker):

```console
$ docker run --rm -it centos:7
[root@637f1d1888f0 /]# printenv
HOSTNAME=637f1d1888f0
TERM=xterm
LS_COLORS=rs=0:di=01;34:ln=01;36:mh=00:pi=40;33:so=01;35:do=01;35:bd=40;33;01:cd=40;33;01:or=40;31;01:mi=01;05;37;41:su=37;41:sg=30;43:ca=30;41:tw=30;42:ow=34;42:st=37;44:ex=01;32:*.tar=01;31:*.tgz=01;31:*.arc=01;31:*.arj=01;31:*.taz=01;31:*.lha=01;31:*.lz4=01;31:*.lzh=01;31:*.lzma=01;31:*.tlz=01;31:*.txz=01;31:*.tzo=01;31:*.t7z=01;31:*.zip=01;31:*.z=01;31:*.Z=01;31:*.dz=01;31:*.gz=01;31:*.lrz=01;31:*.lz=01;31:*.lzo=01;31:*.xz=01;31:*.bz2=01;31:*.bz=01;31:*.tbz=01;31:*.tbz2=01;31:*.tz=01;31:*.deb=01;31:*.rpm=01;31:*.jar=01;31:*.war=01;31:*.ear=01;31:*.sar=01;31:*.rar=01;31:*.alz=01;31:*.ace=01;31:*.zoo=01;31:*.cpio=01;31:*.7z=01;31:*.rz=01;31:*.cab=01;31:*.jpg=01;35:*.jpeg=01;35:*.gif=01;35:*.bmp=01;35:*.pbm=01;35:*.pgm=01;35:*.ppm=01;35:*.tga=01;35:*.xbm=01;35:*.xpm=01;35:*.tif=01;35:*.tiff=01;35:*.png=01;35:*.svg=01;35:*.svgz=01;35:*.mng=01;35:*.pcx=01;35:*.mov=01;35:*.mpg=01;35:*.mpeg=01;35:*.m2v=01;35:*.mkv=01;35:*.webm=01;35:*.ogm=01;35:*.mp4=01;35:*.m4v=01;35:*.mp4v=01;35:*.vob=01;35:*.qt=01;35:*.nuv=01;35:*.wmv=01;35:*.asf=01;35:*.rm=01;35:*.rmvb=01;35:*.flc=01;35:*.avi=01;35:*.fli=01;35:*.flv=01;35:*.gl=01;35:*.dl=01;35:*.xcf=01;35:*.xwd=01;35:*.yuv=01;35:*.cgm=01;35:*.emf=01;35:*.axv=01;35:*.anx=01;35:*.ogv=01;35:*.ogx=01;35:*.aac=01;36:*.au=01;36:*.flac=01;36:*.mid=01;36:*.midi=01;36:*.mka=01;36:*.mp3=01;36:*.mpc=01;36:*.ogg=01;36:*.ra=01;36:*.wav=01;36:*.axa=01;36:*.oga=01;36:*.spx=01;36:*.xspf=01;36:
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
PWD=/
SHLVL=1
HOME=/root
_=/usr/bin/printenv
```

```console
$ docker run --rm -it ubuntu:14.04
root@e7b0dde604db:/# printenv
HOSTNAME=e7b0dde604db
TERM=xterm
LS_COLORS=rs=0:di=01;34:ln=01;36:mh=00:pi=40;33:so=01;35:do=01;35:bd=40;33;01:cd=40;33;01:or=40;31;01:su=37;41:sg=30;43:ca=30;41:tw=30;42:ow=34;42:st=37;44:ex=01;32:*.tar=01;31:*.tgz=01;31:*.arj=01;31:*.taz=01;31:*.lzh=01;31:*.lzma=01;31:*.tlz=01;31:*.txz=01;31:*.zip=01;31:*.z=01;31:*.Z=01;31:*.dz=01;31:*.gz=01;31:*.lz=01;31:*.xz=01;31:*.bz2=01;31:*.bz=01;31:*.tbz=01;31:*.tbz2=01;31:*.tz=01;31:*.deb=01;31:*.rpm=01;31:*.jar=01;31:*.war=01;31:*.ear=01;31:*.sar=01;31:*.rar=01;31:*.ace=01;31:*.zoo=01;31:*.cpio=01;31:*.7z=01;31:*.rz=01;31:*.jpg=01;35:*.jpeg=01;35:*.gif=01;35:*.bmp=01;35:*.pbm=01;35:*.pgm=01;35:*.ppm=01;35:*.tga=01;35:*.xbm=01;35:*.xpm=01;35:*.tif=01;35:*.tiff=01;35:*.png=01;35:*.svg=01;35:*.svgz=01;35:*.mng=01;35:*.pcx=01;35:*.mov=01;35:*.mpg=01;35:*.mpeg=01;35:*.m2v=01;35:*.mkv=01;35:*.webm=01;35:*.ogm=01;35:*.mp4=01;35:*.m4v=01;35:*.mp4v=01;35:*.vob=01;35:*.qt=01;35:*.nuv=01;35:*.wmv=01;35:*.asf=01;35:*.rm=01;35:*.rmvb=01;35:*.flc=01;35:*.avi=01;35:*.fli=01;35:*.flv=01;35:*.gl=01;35:*.dl=01;35:*.xcf=01;35:*.xwd=01;35:*.yuv=01;35:*.cgm=01;35:*.emf=01;35:*.axv=01;35:*.anx=01;35:*.ogv=01;35:*.ogx=01;35:*.aac=00;36:*.au=00;36:*.flac=00;36:*.mid=00;36:*.midi=00;36:*.mka=00;36:*.mp3=00;36:*.mpc=00;36:*.ogg=00;36:*.ra=00;36:*.wav=00;36:*.axa=00;36:*.oga=00;36:*.spx=00;36:*.xspf=00;36:
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
PWD=/
SHLVL=1
HOME=/root
LESSOPEN=| /usr/bin/lesspipe %s
LESSCLOSE=/usr/bin/lesspipe %s %s
_=/usr/bin/printenv
```

Same for my Fedora box.

If we are to push this upstream, we should be aiming to change the OS layer (CentOS), not sti-base, makes sense?

Ultimately, it's hard to tell the impact this would have on end user applications.
@caruccio could you add LC_ALL to your Ruby 1.9 image for now? We can revisit this later if we have demand from other images based on sti-base.

caruccio:
You're right. It makes sense to add this to centos instead sti-base. I confused LC_ALL with LANG, which seams to bet set from attended OS installations.

Anyway, I can go with LC_ALL in my own images. Not a problem.

Thanks for your time.

PR #80, opened 2016-02-29
Expand container-entrypoint args as separate words

### LINKED ISSUE #79: container-entrypoint handling of arguments with spaces
:question: Is this a bug or a feature?

The current implementation of the script:
```bash
cmd="$1"; shift
exec $cmd "$@"
#    ^^^^
```

It handles quoting everywhere except in one place.

It makes a difference when the first argument to the entrypoint has spaces. The current implementation:

- cannot run executables with a space in the name (perhaps not that bad)
- works when multiple arguments are passed as a single argument (to my surprise):

    ```console
$ ./container-entrypoint 'bundle exec rake test'
Could not locate Gemfile
```
- doesn't work in the same case above when the command should not be blindly split on spaces:

    ```console
$ /bin/sh -c "echo 123"
123
$ ./container-entrypoint '/bin/sh -c "echo 123"'
123": -c: line 0: unexpected EOF while looking for matching `"'
123": -c: line 1: syntax error: unexpected end of file
```

This affects image users relying on the entrypoint, for example when they use build hooks.

Shall we correct the quoting, or is it implementing the intended behavior?

rhcarvalho:
Closes #79.

### DIFF
--- a/bin/container-entrypoint
+++ b/bin/container-entrypoint
@@ -1,6 +1,2 @@
 #!/bin/bash
-
-set -eu
-cmd="$1"; shift
-exec $cmd "$@"
-
+exec "$@"


rhcarvalho:
@bparees @mfojtik ptal

mfojtik:
@praiskup PTAL

praiskup:
Wrong - we need `/bin/bash` probably.  Sorry (because we need to take
some `$BASH_*` variables into account.  This probably deserves small comment.

rhcarvalho:
> Wrong - we need /bin/bash probably, because we need to take
> some $BASH_* variables into account.

Yep, the scl enable hack

mfojtik:
@praiskup we need "/bin/sh -i" (which will pickup `ENV` var, but on the other hand produce ugly warning about not having a tty if you run this in non-interactive fashion...) (which is the reason why we should fix 'scl command to provide a way to enable collections globally for docker images ;-)

praiskup:
Yeah - that is something I disagree with (but have to be tolerant ;)).  IMO, SCLs should be
set by openshift -- image consumer.  Image itself should be neutral to allow to build on top
of it without headaches.

praiskup:
s/should be set/should be neabled/

mfojtik:
@praiskup that is a good goal, but openshift consumers wants to have environment similar to what docker language stack provides. Any fluctuation from that setup is a barrier for them to switch to RH images and we don't like annoyed users :-)

rhcarvalho:
@mfojtik can we get this merged?

PR #82, opened 2016-03-15
Fix .pki/nssdb directory permissions

csrwng:
By creating the directory before installing any software and changing its ownership to the default user.

This fixes the issue with builds failing when using aufs as the Docker storage driver.


csrwng:
@bparees ptal

bparees:
lgtm

PR #83, opened 2016-03-15
Add cgroup utility functions

mnagy:
These will be used by https://github.com/openshift/sti-ruby/pull/96

Trello: https://trello.com/c/kvzlecgq/853-5-tune-ruby-based-on-memory-limits-evg

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -83,6 +83,9 @@ COPY bin/container-entrypoint /usr/bin/container-entrypoint
 # Add script to fix directory permissions 
 COPY bin/fix-permissions /usr/bin/fix-permissions
 
+# Add utility for parsing cgroup limits
+COPY bin/cgroup-limits.py /usr/bin/cgroup-limits.py
+
 # Directory with the sources is set as the working directory so all STI scripts
 # can execute relative to this path
 WORKDIR ${HOME}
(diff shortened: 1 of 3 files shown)


mnagy:
@mfojtik @bparees 

@rhcarvalho maybe for python review? :)

mnagy:
The test fail doesn't appear to be my fault.

mfojtik:
LGTM

bparees:
the failure is what i brought up on scrum this morning.  not going to merge this until we can confirm it passes though.

bparees:
[test]

rhcarvalho:
Just so that we share a common context, the decision we made with reading cgroups to limit the container started in a build process is that if we cannot read the limits, we error out and the build will fail.

Possibly here we have different constraints I'd like to learn. What's the guarantee we want to provide for callers of this script?

If we cannot read any one of the cgroup files, we're silently ignoring the error and omitting a env var in the output. What are callers supposed to do when a var is missing?

mfojtik:
@rhcarvalho callers shoud fallback to "default" configuration (iow. use the default values for all tuning flags). @mnagy this should be done for ruby as well.. if we don't have env vars in place, we should skip the tuning.

bparees:
@mfojtik +1

PR #85, opened 2016-03-21
Print warnings to standard error

mnagy:
This will allow us to use the cgroup-limits.py script much easier, with no need for additional warnings:

Example usage:
```
export_vars=$(cgroup-limits.py)
export $export_vars
```

@mfojtik @bparees PTAL

### DIFF
--- a/bin/cgroup-limits.py
+++ b/bin/cgroup-limits.py
@@ -1,5 +1,8 @@
 #!/usr/bin/python
 
+from __future__ import print_function
+import sys
+
 env_vars = {}
 
 
@@ -13,8 +16,11 @@ def read_file(path):
 
 def get_memory_limit():
     limit = read_file('/sys/fs/cgroup/memory/memory.limit_in_bytes')
-    if limit:
-        env_vars['MEMORY_LIMIT_IN_BYTES'] = limit
+    if limit is None:
+        print("Warning: Can't detect memory limit from cgroups",
+              file=sys.stderr)
+        return
+    env_vars['MEMORY_LIMIT_IN_BYTES'] = limit
 
 
 def get_number_of_cores():
@@ -22,6 +28,8 @@ def get_number_of_cores():
 
     line = read_file('/sys/fs/cgroup/cpuset/cpuset.cpus')
     if line is None:
+        print("Warning: Can't detect number of CPU cores from cgroups",
+              file=sys.stderr)
         return
 
     for group in line.split(','):


bin/cgroup-limits.py
```diff
@@ -13,8 +16,10 @@ def read_file(path):
 
 def get_memory_limit():
     limit = read_file('/sys/fs/cgroup/memory/memory.limit_in_bytes')
-    if limit:
-        env_vars['MEMORY_LIMIT_IN_BYTES'] = limit
+    if limit is None:
+        print("Warning: Can't detect memory limit", file=sys.stderr)
```
mfojtik: Warning: Can't read memory limits from ....

We support only one way to detect memory limits, right?

bin/cgroup-limits.py
```diff
@@ -22,6 +27,7 @@ def get_number_of_cores():
 
     line = read_file('/sys/fs/cgroup/cpuset/cpuset.cpus')
     if line is None:
+        print("Warning: Can't detect number of cores", file=sys.stderr)
```
mfojtik: CPU cores

mfojtik:
small nits, but LGTM

mnagy:
Addressed, PTAL

mfojtik:
LGTM

mfojtik:
@mnagy I forgot to mention, can you please document this command (maybe just drop python comment to the file) with things like:

1) what it does, what envs will be returned
2) usage (how I can use it)
3) how to treat errors

bin/cgroup-limits.py
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/python
 
+from __future__ import print_function
```
rhcarvalho: Why try to make Python 2 look like Python 3? :)

mnagy: Just a tiny bit of stuff to make it future proof :) I've used `print()` with parentheses before, but now I need to print something to stderr, which is done differently for the Python 2 print statement and Python 3 print function. We use Python 2.7 and print_function from the future exists since Python 2.6, so this is a safe change :)

There's one other tiny thing in this script that makes this work both for Python 2 and Python 3. It's unnecessary for 2, but would be broken in 3 without it. Bonus points if you can spot it ;)

rhcarvalho: I spotted some things but didn't have the chance to voice it before the first PR was merged :P
Then I let it go in the "this ship has sailed" spirit :cry: 

I don't think aiming at writing code that is cross-version compatible is a goal.
This is running in system Python, which happens to be 2.7. Porting it to 3 should probably be easy for our scripts, when it would be the time (when CentOS/RHEL change their system Python). Many of the things can be ported with automatic tools (2to3).

The thing you refer to is likely [this](https://github.com/mnagy/openshift-base/blob/99b1610fad7224d2efe03547c5114d2f046f50ca/bin/cgroup-limits.py#L36) which makes for ugly Python 2 code :P

PR #86, opened 2016-03-21
Add documentation and usage notes

mnagy:
Add documentation for the script. While looking at pydoc output, I've also realized that I forgot to add a `__name__ == "__main__"` check, so I've added that as well.

@mfojtik PTAL


mfojtik:
LGTM

(would be nice to have some tests for this as well :-) but i'm ok with follow up in future.

PR #87, opened 2016-03-23
Remove .py suffix from cgroup-limits script

mnagy:
@rhcarvalho @bparees


rhcarvalho:
perhaps we could do this in 2 steps to avoid breaking stuff:

1. Copy the file with the new name
2. Do work on all images that depend on the old name to use the new
3. Remove the old file

bparees:
i think the base test actually builds+tests every downstream image, so we should be good.  merging.

PR #90, opened 2016-04-13
Simplify cgroup script, add NO_MEMORY_LIMIT variable

mnagy:
@mfojtik @bparees

### DIFF
--- a/bin/cgroup-limits
+++ b/bin/cgroup-limits
@@ -23,12 +23,13 @@ Variables currently supported:
     NUMBER_OF_CORES
         Number of detected CPU cores that can be used. This value is
         calculated from /sys/fs/cgroup/cpuset/cpuset.cpus
+    NO_MEMORY_LIMIT
+        Set to "true" if MEMORY_LIMIT_IN_BYTES is so high that the caller
+        can act as if no memory limit was set. Undefined otherwise.
 """
 
 from __future__ import print_function
 import sys
-
-env_vars = {}
 
 
 def _read_file(path):
@@ -45,11 +46,11 @@ def get_memory_limit():
     """
 
     limit = _read_file('/sys/fs/cgroup/memory/memory.limit_in_bytes')
-    if limit is None:
+    if limit is None or not limit.isdigit():
         print("Warning: Can't detect memory limit from cgroups",
               file=sys.stderr)
-        return
-    env_vars['MEMORY_LIMIT_IN_BYTES'] = limit
+        return None
+    return int(limit)
 
 
 def get_number_of_cores():
@@ -63,7 +64,7 @@ def get_number_of_cores():
     if line is None:
         print("Warning: Can't detect number of CPU cores from cgroups",
               file=sys.stderr)
-        return
+        return None
 
     for group in line.split(','):
         core_ids = list(map(int, group.split('-')))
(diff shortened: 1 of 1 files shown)


bin/cgroup-limits
```diff
@@ -72,12 +73,20 @@ def get_number_of_cores():
...
-    get_number_of_cores()
+    env_vars = {
+        "MAX_MEMORY_LIMIT_IN_BYTES": 9223372036854775807,
+        "MEMORY_LIMIT_IN_BYTES": get_memory_limit(),
+        "NUMBER_OF_CORES": get_number_of_cores()
+    }
+
+    env_vars = {k: v for k, v in env_vars.items() if v is not None}
+
+    if env_vars.get("MEMORY_LIMIT_IN_BYTES", 0) > 92233720368547:
```
bparees: this is the value that s2i will set the container limit to if cgroups reports no limit, so this code needs to check if the value is >= that value, not just >

bparees:
@mnagy one nit and can you also list the other PRs that need to go in simultaneously with this one?  (is it just the postgres one?)

bin/cgroup-limits
mnagy: I could have sworn I had that there before. Nice catch.

mnagy:
@bparees I don't think any need to go simultaneously, because this change doesn't break any previous behaviour, it just adds the new variable. We might want to change the behaviour somewhere. @mfojtik will want to update the script that goes into postgres, @php-coder will need the new script for his mysql card and I'm not sure if somebody is doing mongo.

Additionally, I know @mfojtik needed to know if there is no limit set in one of his PR. For sti-ruby, there'll be no change required.

mnagy:
PTAL

bparees:
>  I don't think any need to go simultaneously, because this change doesn't break any previous behaviour, it just adds the new variable. 

yup, i was overthinking it. thanks.

>  For sti-ruby, there'll be no change required.

why's that?

lgtm but would like @mfojtik to review since he's actually used this stuff.

bin/cgroup-limits
```diff
@@ -72,12 +73,20 @@ def get_number_of_cores():
...
-    get_number_of_cores()
+    env_vars = {
+        "MAX_MEMORY_LIMIT_IN_BYTES": 9223372036854775807,
+        "MEMORY_LIMIT_IN_BYTES": get_memory_limit(),
+        "NUMBER_OF_CORES": get_number_of_cores()
+    }
+
+    env_vars = {k: v for k, v in env_vars.items() if v is not None}
+
+    if env_vars.get("MEMORY_LIMIT_IN_BYTES", 0) >= 92233720368547:
```
php-coder: `92233720368547` is the same as `9223372036854775807`? Maybe we can use `env_vars.get("MAX_MEMORY_LIMIT_IN_BYTES")` here?

mnagy:
It's because we first calculate the number of processes we want to use (using the number of cores) and then we use the memory limit to calculate the maximum number of processes and pick the lower value as the actuall number. See here: https://github.com/openshift/sti-ruby/blob/master/2.0/contrib/etc/puma.cfg#L23

bin/cgroup-limits
mnagy: No, I'm comparing the actual limit. I used to get that number by calling get_memory_limit() but changed it because it read the file two times (and also issued two warnings if it couldn't get the limit).

php-coder: Hm... I mean something like that:
```python
if env_vars.get("MEMORY_LIMIT_IN_BYTES", 0) >= env_vars.get("MAX_MEMORY_LIMIT_IN_BYTES"):
```

bparees: > 92233720368547 is the same as 9223372036854775807?

no.  The first one represents a practical upper bound that implies no limit, the second one represents the actual value cgroups reports when there is no limit.

bparees:
> It's because we first calculate the number of processes we want to use (using the number of cores) and then we use the memory limit to calculate the maximum number of processes and pick the lower value as the actuall number. See here: https://github.com/openshift/sti-ruby/blob/master/2.0/contrib/etc/puma.cfg#L23

so put another way, the ruby image doesn't have any specific logic for "if no limit is set"?

mfojtik:
LGTM, will leave @bparees to merge :-)

mnagy:
@bparees correct.

bparees:
will wait on tests to pass :)

PR #92, opened 2016-04-24
Added extras repo

jason-callaway:
Had to add rhel-7-server-extras-rpms for bsdtar and gd-devel


bparees:
is this needed to build in osbs?
can you squash this down to one commit?

hhorak:
`bsdtar` is a subpackage of `libarchive` and so it should be available in `optional` channel. I see no reason to enable extra channel.

PR #96, opened 2016-06-13
fix recursive handling of links

gabemontero:
@bparees PTAL

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -1,16 +1,19 @@
 #!/bin/bash -e
 # This script is used to build, test and squash the OpenShift Docker images.
 #
-# $1 - Specifies distribution - "rhel7" or "centos7"
+# OS - Specifies distribution - "rhel7" or "centos7"
+# BASE_IMAGE_DIR - Specifies the base image name - "base"
 # TEST_MODE - If set, build a candidate image and test it
 # TAG_ON_SUCCESS - If set, tested image will be re-tagged as a non-candidate
 #                  image, if the tests pass.
 
-OS=$1
-
 DOCKERFILE_PATH=""
-BASE_DIR_NAME=$(echo $(basename `pwd`) | sed -e 's/-[0-9]*$//g')
-BASE_IMAGE_NAME="openshift/${BASE_DIR_NAME#s2i-}"
+test -z "$BASE_IMAGE_NAME" && {
+    BASE_DIR_NAME=$(echo $(basename `pwd`) | sed -e 's/-[0-9]*$//g')
+    BASE_IMAGE_NAME="${BASE_DIR_NAME#s2i-}"
+}
+
+NAMESPACE="openshift/"
 
 # Cleanup the temporary Dockerfile created by docker build with version
 trap "rm -f ${DOCKERFILE_PATH}.version" SIGINT SIGQUIT EXIT
@@ -40,7 +43,7 @@ function squash {
   ${HOME}/.local/bin/docker-scripts squash -f $base ${IMAGE_NAME}
 }
 
-IMAGE_NAME="${BASE_IMAGE_NAME}-${OS}"
+IMAGE_NAME="${NAMESPACE}${BASE_IMAGE_NAME}-${OS}"
 
 if [[ -v TEST_MODE ]]; then
   IMAGE_NAME+="-candidate"
--- a/Makefile
+++ b/Makefile
@@ -6,10 +6,21 @@ else
 	OS := centos7
 endif
 
+BASE_IMAGE_NAME = "base"
+
+script_env = \
+	SKIP_SQUASH=$(SKIP_SQUASH)                      \
+	VERSIONS="$(VERSIONS)"                          \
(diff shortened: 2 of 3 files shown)


bparees:
the test job for this repo may need some work due to the repo renaming (sti-base to s2i-base), not sure
[test]

bparees:
you may also need to rebase if you haven't already to make sure we're testing current code.

gabemontero:
@bparees i rebased before submitting the pull

gabemontero:
I think this is the failing error:

```
+ vagrant ssh -c 'docker tag openshift/base-rhel7-candidate openshift/base-rhel7'

Error response from daemon: no such id: openshift/base-rhel7-candidate
```

bparees:
Yeah but I think that is the correct name so i'm not sure why it isn't
there.  Will have to dig into the repo scripts and vagrant plugin.

Ben Parees | OpenShift

gabemontero:
Still digging, but the entry message from build.sh is:
`-> Building openshift/sti-base-rhel7-candidate ...`

So there is an `sti-` prefix that is not present in the `docker tag openshift/base-rhel7-candidate openshift/base-rhel7`

gabemontero:
BASE_IMAGE_NAME="openshift/${BASE_DIR_NAME#s2i-}" appears to strip "s2i-" when I run it and the preceding lines in build.sh in my "openshift/s2i-base" directory, resulting in "openshift/base" as the later `docker tag ..` seems to expect.

This would imply to me that on jenkins, the repo must be getting extracted to the old "openshift/sti-base" directory.

Further up the log, I see this:
`+ vagrant test-origin-image --image base --source openshift/sti-base --ref 78f31e7820e3b0343ff21134d5478f7773c42857`

which corresponds to this in the job definition:

`vagrant test-origin-image --image base --source openshift/$JOB_NAME --ref $ghprbActualCommit`

Where JOB_NAME is `sti-base`.

So we either change the job name to s2i-base, or we change to job config to not use JOB_NAME in the vagrant checkout.  The latter seems less invasive, I'll wait for @bparees to see this and chime in before pulling the trigger.

bparees:
Makefile for s2i-base should be updated to define the builder_image like our other repos do:
https://github.com/sclorg/s2i-ruby-container/blob/master/Makefile#L2

and hack/build.sh should be updated to expect that value first, like we do here:
https://github.com/sclorg/s2i-ruby-container/blob/master/hack/build.sh#L21-L24

you can just do that in this PR to get things rolling.

gabemontero:
Just setting an env var in the Makefile was not proving sufficient for my testing (both local and the PR).  I had to pass the value as a param into build.sh.

bparees:
You need to do this:

https://github.com/sclorg/s2i-ruby-container/blob/master/hack/common.mk#L16

And the base image name value should be just "base".

gabemontero:
latest updates pushed @bparees

hack/build.sh
```diff
@@ -1,16 +1,19 @@
...
 #                  image, if the tests pass.
 
-OS=$1
-
 DOCKERFILE_PATH=""
-BASE_DIR_NAME=$(echo $(basename `pwd`) | sed -e 's/-[0-9]*$//g')
-BASE_IMAGE_NAME="openshift/${BASE_DIR_NAME#s2i-}"
+test -z "$BASE_IMAGE_NAME" && {
+    BASE_DIR_NAME=$(echo $(basename `pwd`) | sed -e 's/-[0-9]*$//g')
+    BASE_IMAGE_NAME="openshift/${BASE_DIR_NAME#s2i-}"
```
bparees: you still don't need openshift here, it's prepended below.  (though this logic shouldn't be executed now, normally, so it's not the current issue)

bparees:
current failure looks like a flake (the base image tester builds every downstream image and tests it as part of testing the change, so it can be finicky)
[test]

hack/build.sh
gabemontero: removed "openshift/" and pushed ... let's see what the next PR test run reveals

gabemontero:
On Tue, Jun 14, 2016 at 3:17 AM, Ben Parees <notifications@github.com>
wrote:

> current failure looks like a flake (the base image tester builds every
> downstream image and tests it as part of testing the change, so it can be
> finicky)
> [test]
>

Yeah I noticed that with a couple of earlier runs that got past the
original problem.

bparees:
lgtm pending a test pass.

gabemontero:
And we have success in https://ci.openshift.redhat.com/jenkins/job/sti-base/161/

```
Sending e-mails to: jhadvig@redhat.com mfojtik@redhat.com

Setting status of 921c12de429051680114a60880dd10d6e123cdfd to SUCCESS with url https://ci.openshift.redhat.com/jenkins/job/sti-base/160/ and message: ' '
Finished: SUCCESS

```

gabemontero:
Oops ... I see it started https://ci.openshift.redhat.com/jenkins/job/sti-base/161/

PR #97, opened 2016-06-13
[DO_NOT_MERGE] TEST: IGNORE

bparees:


bparees:
@gabemontero we can see what happens over here.

gabemontero:
yep @bparees same error as with https://github.com/openshift/s2i-base/pull/96

bparees:
K,  will look into it

gabemontero:
@bparees see https://github.com/openshift/s2i-base/pull/96 ... got some triage and proposed solutions.

PR #101, opened 2016-08-12
Add bzip2 to Dockerfiles

kasicka:
Customer can't use the image, because the bzip2 package is missing.
https://bugzilla.redhat.com/show_bug.cgi?id=1364546
https://github.com/sclorg/s2i-nodejs-container/pull/114

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,6 +35,7 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
   INSTALL_PKGS="autoconf \
   automake \
   bsdtar \
+  bzip2 \
   epel-release \
   findutils \
   gcc-c++ \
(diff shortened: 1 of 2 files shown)


bparees:
lgtm

bparees:
you'll need to add this to the sclorg base image also, right?
@hhorak

hhorak:
@bparees We still keep using this repo for all centos-based images in sclorg/, so it will be picked once nodejs images is rebuild next time. Should we just create a dummy commit to make it happen or is there a better way?
About forking s2i base generally, I'm not sure whether having a fork of s2i base would bring any benefits so far, we just have a fork downstream for RHEL-based images, because of slightly different release schedule. And I think we can move this repo to sclorg/ once it is no longer needed for older version of the images and thus I expect you won't be willing to maintain it anymore. Of course I'm open to other suggestions.

bparees:
@hhorak since the base image changed, all the downstream centos images will be rebuilt and pushed to dockerhub automatically by our jenkins infrastructure.

bparees:
and yes, once the images we maintain go EOL we can move this repo.

PR #102, opened 2016-10-06
Added cccp.yml for the Container Pipeline

mohammedzee1000:
This is for the purpose of building this container on the
centos container pipeline.


hhorak:
It does not influence image functionality, so no need to test. LGTM.

PR #104, opened 2016-10-12
remove the epel repo

bparees:


kanedafromparis:
I wonder why you remove epel repo ?

bparees:
because we don't need it to build the image.

kanedafromparis:
ok, thanks

PR #109, opened 2017-01-17
Optimization fix-permissions

### LINKED ISSUE #108: fix-permission optimization
fix-permission is very slow.
instead of
```
#!/bin/sh
# Fix permissions on the given directory to allow group read/write of 
# regular files and execute of directories.
find $1 -exec chgrp 0 {} \;
find $1 -exec chmod g+rw {} \;
find $1 -type d -exec chmod g+x {} +
```

I tried with
```
#!/bin/sh
# Fix permissions on the given directory to allow group read/write of
# regular files and execute of directories.
chgrp -R 0 $1;
chmod -R g+rw $1;
find $1 -type d -exec chmod g+x {} +
```
it is faster at least 100x
both scripts produce the same result.

cata86:
fixes #108

### DIFF
--- a/bin/fix-permissions
+++ b/bin/fix-permissions
@@ -2,5 +2,7 @@
 # Fix permissions on the given directory to allow group read/write of
 # regular files and execute of directories.
 chgrp -R 0 $1;
+find -L $1 -xtype l -exec chgrp 0 {} \;
 chmod -R g+rw $1;
+find -L $1 -xtype l -exec chmod g+rw {} \;
 find $1 -type d -exec chmod g+x {} +


bin/fix-permissions
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
-# Fix permissions on the given directory to allow group read/write of 
+# Fix permissions on the given directory to allow group read/write of
 # regular files and execute of directories.
-find $1 -exec chgrp 0 {} \;
-find $1 -exec chmod g+rw {} \;
+chgrp -R 0 $1;
+chmod -R g+rw $1;
```
bparees: this does not have the same behavior as find when the directory contains symlinked files.

bparees:
[test]

PR #110, opened 2017-02-09
Updated hack/build.sh to use the newer docker-squash tool

Djelibeybi:
I was trying to build an Oracle Linux base image and the squashing never worked. Turns out the old version of docker-scripts no longer works with Docker 1.12 or 1.13. I've switched to using docker-squash instead, which works just fine.

The command essentially squashes the image back to the specified tag.

Signed-off-by: Avi Miller <avi.miller@oracle.com>

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -26,9 +26,11 @@ function docker_build_with_version {
   cp ${DOCKERFILE_PATH} "${DOCKERFILE_PATH}.version"
   git_version=$(git rev-parse --short HEAD)
   echo "LABEL io.openshift.builder-base-version=\"${git_version}\"" >> "${dockerfile}.version"
-  docker build -t ${IMAGE_NAME} -f "${dockerfile}.version" .
   if [[ "${SKIP_SQUASH}" -ne "1" ]]; then
+    docker build -t ${IMAGE_NAME}-unsquashed -f "${dockerfile}.version" .
     squash "${dockerfile}.version"
+  else
+    docker build -t ${IMAGE_NAME} -f "${dockerfile}.version" .
   fi
   rm -f "${DOCKERFILE_PATH}.version"
 }
@@ -38,7 +40,7 @@ function docker_build_with_version {
 function squash {
   easy_install -q --user docker-squash
   base=$(awk '/^FROM/{print $2}' $1)
-  ${HOME}/.local/bin/docker-squash -f $base -t ${IMAGE_NAME} ${IMAGE_NAME}
+  ${HOME}/.local/bin/docker-squash -c -f $base -t ${IMAGE_NAME} ${IMAGE_NAME}-unsquashed
 }
 
 IMAGE_NAME="${NAMESPACE}${BASE_IMAGE_NAME}-${OS}"


bparees:
@hhorak personally i'd be supportive of stripping the squashing logic out entirely at this point(it's been problematic in various ways over the years), but it's your call.

Djelibeybi:
@bparees @hhorak Docker 1.13 (Experimental) has squashing capabilities so perhaps strip it out now and then restore it once the Docker client itself can do the squashing?

bparees:
@Djelibeybi yeah that also sounds reasonable.

hhorak:
[test]

hhorak:
I'm not aware of problems with squashing, but if that is really problematic and not working with newer client, I'd be fine with either way -- turning off the squashing or changing the tool for squashing. But I'd like to understand the issue first..
@omron93 , what do you think, are you aware of the issues currently? I see we already have docker 1.12 on the builder machines.

omron93:
> I'm not aware of problems with squashing, but if that is really problematic and not working with newer client, I'd be fine with either way -- turning off the squashing or changing the tool for squashing. But I'd like to understand the issue first..

Yes, it is true. It is not working for s2i-base. Other images (I hope all) are using `docker-squash==1.0.1` (1.0.5 is the latest). So +1 for this PR.

@Djelibeybi Please, could you describe more the second commit "A little more logic to make sure the squash process properly removes the unsquashed image."? (because it creates another difference according to scripts of other sclorg/ images)

> -- turning off the squashing or changing the tool for squashing. But I'd like to understand the issue first..
@omron93 , what do you think, are you aware of the issues currently? I see we already have docker 1.12 on the builder machines.

@hhorak I don't know. Only I remember Ben was mentioning "i've seen cases where squashing changes behavior" [1] @bparees Do you have more information please?

Currently squashing is not used or tested in CI (because we are not releasing squashed images and tests would run two times longer). @hhorak Should we test it?

Squashing using docker is currently **experimental** (added in 1.13).
So I vote for using it after it gets stable (currently is is possible to add custom parameters for `docker build` - `BUILD_OPTIONS` variable), documenting how to do it with `BUILD_OPTIONS` and removing code for squashing.

[1] https://github.com/sclorg/mariadb-container/pull/9#issuecomment-238559171

omron93:
[test]

bparees:
>  Only I remember Ben was mentioning "i've seen cases where squashing changes behavior" [1] @bparees Do you have more information please?

we had issues w/ file permissions getting set incorrectly, as i recall.

Djelibeybi:
@omron93 the `docker-squash` tool has slightly different mechanics to the old tool, so it was necessary to create a "before" image that can get squashed.

hhorak:
> Currently squashing is not used or tested in CI (because we are not releasing squashed images and tests would run two times longer). @hhorak Should we test it?

Twice longer because we would run the test once for squashed and once for unsquashed image?
I think we should ideally test what is later delivered, so if we don't squash centos images, but we do squash rhel images, we might set squashing before testing for RHEL testing only. Does that make sense?

omron93:
> the docker-squash tool has slightly different mechanics to the old tool, so it was necessary to create a "before" image that can get squashed.

@Djelibeybi You are right. Is there a benefit of having `...-unsquashed` image instead of working with `IMAGE_NAME` in all steps?

> Twice longer because we would run the test once for squashed and once for unsquashed image?

Exactly.

> I think we should ideally test what is later delivered, so if we don't squash centos images, but we do squash rhel images, we might set squashing before testing for RHEL testing only. Does that make sense?

@hhorak OK. So for CentOS based images test will be run with non-squashed image and for RHEL based images tests will be using squashed images. I will set it up...

pkubatrh:
>I think we should ideally test what is later delivered, so if we don't squash centos images, but we do squash rhel images, we might set squashing before testing for RHEL testing only. 

I agree we should test the squashed images for RHEL (via docker-squash) if that is what we are shipping.

>Is there a benefit of having ...-unsquashed image instead of working with IMAGE_NAME in all steps?

The only benefit I see is being able to use the `-c` option to remove the unsquashed image that would otherwise be leftover.

hhorak:
[test]

hhorak:
We use container-common-scripts repo and particularly this script for building: https://github.com/sclorg/container-common-scripts/blob/master/build.sh 
At this point, we don't install docker-squash at all and just instruct the builder to acquire this tool. It makes sense to me this way.

praiskup:
Btw., the the latest build.sh script doesn't squash centos7 image anymore (only rhel7).

PR #111, opened 2017-02-09
Add ol7 base

Djelibeybi:
This PR adds the ability to build an s2i image using Oracle Linux 7 as the base.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,11 +41,16 @@ $ make build
 
 * **Oracle Linux 7 base image**
 
+This image is available on the Docker Hub. To download it run:
+
+```console
+docker pull oracle/s2i-base-oraclelinux7
+```
+
 To build an Oracle Linux 7 based image from scratch run:
 
 ```
-$ git clone https://github.com/openshift/s2i-base.git
-$ cd s2i-base
+$ git clone https://github.com/sclorg/s2i-base-container.git
 $ make build TARGET=ol7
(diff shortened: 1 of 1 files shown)


bparees:
@hhorak this one is above my paygrade :)

hhorak:
@Djelibeybi Thanks for your PR, but I'm not convinced we should include more distros, because it would significantly increase maintenance cost. The cost is already not small even now with 2 or 3 variants, and 2-4 versions of the particular image (python, postgresql, mongodb), and adding any distro more would mean to do changes on even more places.

Djelibeybi:
@hhorak I guess I see Oracle Linux as the middle ground between CentOS (Full community, no commercial support) and RHEL (requires commercial support).

Oracle Linux is free to download, distribute and use with free updates provided, but is also eligible for commercial support from Oracle if required. 

However, this is your repo, so your rules. :)

hhorak:
I see you point, although with so many derivatives (https://upload.wikimedia.org/wikipedia/commons/a/a3/Redhat_family_tree_11-06.png, not saying all are on the same level as Oracle Linux) the maintenance gets quickly worse and we already fight with this today.. Anyway, I'll bring this idea up internally to see more PoVs and let know here (hopefully soon).

Djelibeybi:
This is outdated. I can submit a new one.

PR #113, opened 2017-03-17
Allow fix-permissions to fail without failing a build

csrwng:
If some files can't be updated because the current user doesn't own them or some other reason, the entire build should not fail. This will allow secrets to be injected inside the source tree.


csrwng:
@openshift/dev-ex ptal

bparees:
[test]

PR #118, opened 2017-04-17
ensure proper rpms installed

tchughesiv:


Dockerfile.rhel7
```diff
@@ -1,5 +1,5 @@
 # This image is the base image for all OpenShift v3 language Docker images.
-FROM rhel7
+FROM registry.access.redhat.com/rhel7
```
bparees: this value("rhel7") is presently a necessary artifact of how we do our internal builds of this image.

tchughesiv: gotcha, reverted w/ https://github.com/sclorg/s2i-base-container/pull/118/commits/ef46e0ebb849d5a79cf26b4ff0daa7d441178d78

bparees:
thanks, i'll look into changing our internal behavior so we can make this align in the future.

[test]

bparees:
sorry I was thinking this was on a different repo.  @hhorak you guys ok w/ these changes to the enabled repo list?

omron93:
+1

Other images using `s2i-base` as a base image would benefit from this too.

hhorak:
Makes sense to me as well. Merging.

PR #114, opened 2017-03-20
Change default image name

omron93:
This image is built as `centos/s2i-base-centos7` (https://hub.docker.com/r/centos/s2i-base-centos7/). So changing default name for the image.

@bparees Please take a look and merge.

Issue #112, opened 2017-02-16
Warning messages when fix-permissions run on a symlink.


GrahamDumpleton:
In the Python S2I builder, the directories which ``fix-permissions`` is run on can include symlinks which point to somewhere outside of the directory used for the application. The directories/files being pointed at can be owned by root.

The result of this is that ``fix-permissions`` generates lots of warnings of type:

```
chgrp: changing group of '/opt/app-root/include/python2.7': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/UserDict.py': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/_abcoll.py': Operation not permitted
...
chmod: changing permissions of '/opt/app-root/include/python2.7': Operation not permitted
chmod: changing permissions of '/opt/app-root/lib/python2.7/UserDict.py': Operation not permitted
chmod: changing permissions of '/opt/app-root/lib/python2.7/_abcoll.py': Operation not permitted
chmod: changing permissions of '/opt/app-root/lib/python2.7/_weakrefset.py': Operation not permitted
```

This is because ``chgrp`` and ``chmod`` act on what the symlink points to and not the symlink and thus fail.

This was warned about in:

* https://github.com/sclorg/s2i-python-container/pull/159

but nothing was done about it before PR was merged.

Bumping this to base image for addressing.

The only known workaround would be to use the ``-f`` option with ``chmod`` and ``chgrp`` so it ignores the failures. It is not known though whether that would mask any other problems.

Only other solution would be stop trying to use shell commands for this and write a custom C/go program which does a more intelligent traversal of the directory hierarchy to fix up permissions and ownership.

GrahamDumpleton:
The ``fix-permissions`` as written suffers other problems as well. A summary of all the problems are as follows.

1: Due to:

```
find -L $1 -xtype l -exec chgrp 0 {} \;
find -L $1 -xtype l -exec chmod g+rw {} \;
```

the script is trying to update the permissions of directories or files which are the target of symlinks. This is what causes the errors such as:

```
chgrp: changing group of '/opt/app-root/include/python2.7': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/UserDict.py': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/_abcoll.py': Operation not permitted
...
chmod: changing permissions of '/opt/app-root/include/python2.7': Operation not permitted
[...]

PR #120, opened 2017-04-24
Avoid errors when the process does not have enough permissions

hhorak:
Sometimes there is not enough permissions to change attributes by `fix-permissions`. It was for example reported in sclorg/mariadb-container#32. This change should make these warnings disappeared.

### DIFF
--- a/bin/fix-permissions
+++ b/bin/fix-permissions
@@ -3,10 +3,18 @@
 # Allow this script to fail without failing a build
 set +e
 
-# Fix permissions on the given directory to allow group read/write of
+# Fix permissions on the given directory or file to allow group read/write of
 # regular files and execute of directories.
-chgrp -R 0 $1;
-find -L $1 -xtype l -exec chgrp 0 {} \;
-chmod -R g+rw $1;
-find -L $1 -xtype l -exec chmod g+rw {} \;
-find $1 -type d -exec chmod g+x {} +
+
+# If argument does not exist, script will still exit with 0,
+# but at least we'll see something went wrong in the log
+if ! [ -e "$1" ] ; then
+  echo "ERROR: File or directory $1 does not exist." >&2
+fi
+
+# Using --quiet to avoid error messages when there are not enough permissions
+chgrp -R --quiet 0 "$1";
+find -L "$1" -xtype l -exec chgrp --quiet 0 {} \;
+chmod -R --quiet g+rw "$1";
+find -L "$1" -xtype l -exec chmod --quiet g+rw {} \;
+find "$1" -type d -exec chmod --quiet g+x {} +
--- a/bin/fix-permissions
+++ b/bin/fix-permissions
@@ -13,8 +13,7 @@ if ! [ -e "$1" ] ; then
 fi
 
 # Using --quiet to avoid error messages when there are not enough permissions
-chgrp -R --quiet 0 "$1";
-find -L "$1" -xtype l -exec chgrp --quiet 0 {} \;
-chmod -R --quiet g+rw "$1";
-find -L "$1" -xtype l -exec chmod --quiet g+rw {} \;
-find "$1" -type d -exec chmod --quiet g+x {} +
+find -L "$1" \! -gid 0 -exec chgrp 0 {} +
+find -L "$1" \! -perm /g+rw -exec chmod g+rw {} +
+find -L "$1" -perm /u+x -a \! -perm /g+x -exec chmod g+x {} +
+find -L "$1" -type d \! -perm /g+x -exec chmod g+x {} +
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -13,7 +13,11 @@ test -z "$BASE_IMAGE_NAME" && {
     BASE_IMAGE_NAME="${BASE_DIR_NAME#s2i-}"
 }
 
-NAMESPACE="centos/"
+if [ "${OS}" == "rhel7" ]; then
+    NAMESPACE="rhscl/"
+else
+    NAMESPACE="centos/"
+fi
 
 # Cleanup the temporary Dockerfile created by docker build with version
 trap "rm -f ${DOCKERFILE_PATH}.version" SIGINT SIGQUIT EXIT


pkubatrh:
I wonder, is there a chance we might mask some future issues when suppressing warnings and errors while running `fix-permissions`? Nothing comes to my mind right now but in general it seems like a bad idea.

There is also https://github.com/sclorg/s2i-base-container/issues/112 that might get influenced by this change.

How about making `fix-permissions` take the GID via an argument (would default to 0 so we do not have to change all existing calls) instead? That would get rid of the warnings while additionally making sure the process has the correct access permissions for the files.

hhorak:
> I wonder, is there a chance we might mask some future issues when suppressing warnings and errors while running fix-permissions? Nothing comes to my mind right now but in general it seems like a bad idea.

I share the general worry, but thinking more about it it looks to me like the only real issue that we might mask is when some directory we pass to the script does not exist. That's why there is a check `if ! [ -e "$1" ] ; then`.. But I agree that we never can be sure.

> How about making fix-permissions take the GID via an argument (would default to 0 so we do not have to change all existing calls) instead? That would get rid of the warnings while additionally making sure the process has the correct access permissions for the files.

I don't think it would make all errors disappeared, because symlinks mentioned in #112 that point to root-owned files (correctly read-only) would still produce warnings/errors IMO.. the reason of not having enough permissions is different there.

omron93:
Problem is that the user is not in `root` group. So it is not possible to `chgrp` to such group.

@hhorak For example for mongodb-container - adding `usermod -a -G root mongodb` in Dockerfile solves this issue.

pkubatrh:
>For example for mongodb-container - adding usermod -a -G root mongodb in Dockerfile solves this issue.

Is this something that we want to have/is expected of containers running in Openshift?

omron93:
>  How about making fix-permissions take the GID via an argument (would default to 0 so we do not have to change all existing calls) instead? That would get rid of the warnings while additionally making sure the process has the correct access permissions for the files.

Copies of `fix-permissins` scripts in mongodb-container [1] and postgresql-container [2] also change owner to default image user. This is needed for running using "pure" docker where default group isn't `root`.
On the other hand changing owner makes sense only in Dockerfile running under root. So +1 taking arguments to fix-permissions. 

[1] https://github.com/sclorg/mongodb-container/blob/master/3.2/root/usr/libexec/fix-permissions
[2] https://github.com/sclorg/postgresql-container/blob/master/9.5/root/usr/libexec/fix-permissions

> I don't think it would make all errors disappeared, because symlinks mentioned in #112 that point to root-owned files (correctly read-only) would still produce warnings/errors IMO.. the reason of not having enough permissions is different there.

We should not change file attributes of similar files. Warning makes sense in this case IMPOV.

omron93:
> Is this something that we want to have/is expected of containers running in Openshift?

In OpenShift all containers are running under `root` GID and random UID. That is why all files should be accessible by `root` group. @bparees Is it right?

bparees:
> In OpenShift all containers are running under root GID and random UID. That is why all files should be accessible by root group. @bparees Is it right?

sort of.  The assemble script actually runs as the default user for the image, and if the default user for the image has specifically assigned groups, then those are the groups they will have, not necessarily the root group.

But if the user is not listed in /etc/groups, then yes they get the root group.

omron93:
> But if the user is not listed in /etc/groups, then yes they get the root group.

@bparees So when running container in OpenShift with "random" UID gets always root GID, or not? Is it possible that "random" UIG will be already listed in `/etc/passwd`?

Is it problem to add default user to root group from OpenShift POV ?

bparees:
> @bparees So when running container in OpenShift with "random" UID gets always root GID, or not? Is it possible that "random" UIG will be already listed in /etc/passwd?

it's theoretically possible the user could be listed (and thus not have root group permissions) but pretty unlikely.  the assigned/random uid range starts at like 10000 or something.  i've never seen an image that defines user 10000 in their /etc/passwd and /etc/groups files.

> Is it problem to add default user to root group from OpenShift POV ?

no i don't think that should be a problem (but again you don't need to add the default user explicitly as long as the default user isn't otherwise already listed.  for most of our images we use the default user 1001, which is why this works ok for us).

omron93:
> no i don't think that should be a problem (but again you don't need to add the default user explicitly as long as the default user isn't otherwise already listed. for most of our images we use the default user 1001, which is why this works ok for us).

In database images default user is different. So without random UID it causes problems.

So solution it to also use 1001 user in database images or to add current default users to root group. @hhorak What to choose?

hhorak:
I prefer to add the default users to the root group, since this issue does not seem worth changing the default user. So, it should remove the need to hide errors in case of `chgrp` calls, but there is still the issue with files owned by root (https://github.com/sclorg/s2i-python-container/issues/178#issue-203747032) where we see error messages when symlinks point to them, which we still should mitigate somehow.

hhorak:
A fixed solution with suggestions from  https://github.com/sclorg/s2i-base-container/issues/112#issuecomment-293217175 is available.

omron93:
[test]

pkubatrh:
`Operation not permitted` warnings still present in RHEL7 CI as it seems to be testing s2i-* images built upon the wrong version of s2i-base (the PR version is tagged as `centos/s2i-base-rhel7` instead of the expected `rhscl/s2i-base-rhel7`).
Would like to get this fixed and run through the RHEL CI properly before merging.

omron93:
> Operation not permitted warnings still present in RHEL7 CI as it seems to be testing s2i-* images built upon the wrong version of s2i-base (the PR version is tagged as centos/s2i-base-rhel7 instead of the expected rhscl/s2i-base-rhel7).
Would like to get this fixed and run through the RHEL CI properly before merging.

You are right! Build scripts for this image does not support `centos/` vs `rhscl/` tagging... It used `openshift/` namespace for both versions, but after https://github.com/sclorg/s2i-base-container/pull/114 namespace is always `centos/` instead of `openshift/`

I suggest to fix this by using https://github.com/sclorg/container-common-scripts (for example by accepting https://github.com/sclorg/s2i-base-container/pull/121)

pkubatrh:
@hhorak Please rebase to be able run tests including https://github.com/sclorg/s2i-base-container/pull/122

PR #123, opened 2017-05-17
Avoid errors when the process does not have enough permissions to change attributes

hhorak:
I tried to rebase the original #120 by an innovative way, but it ended bad. This PR is the same as #120.

PR #124, opened 2017-05-24
Always succeed when running fix-permissions

csrwng:
Exits with success from fix-permissions script ignoring any errors that may have occurred.
https://bugzilla.redhat.com/show_bug.cgi?id=1433027


csrwng:
@bparees ptal
The fix-permissions script would still fail if the last statement in the script returned a non-zero exit code. This makes sure that a 0 is always returned. Tested with example from bz.

bparees:
lgtm
[test]

bparees:
@hhorak can you pick someone to review/merge this?

hhorak:
Looks good to me, merging. There is also another PR #123 (basically same as #120 which I screw rebase for), where I'll include this change (`exit 0`) as well, but since you're looking at this @csrwng, would you like to comment there as well to see whether it makes sense to you?

PR #125, opened 2017-06-20
fix fault in fix-permissions where group perms were not set correctly

jim-minter:
Currently this line means: find all files where group read is not set AND group write is not set, and chmod them g+rw.
I think it should be: find all files where group read is not set OR group write is not set, and chmod them g+rw.

Otherwise files which are already g+r do not become g+rw.

This broke the origin extended image ecosystem test.

@hhorak

### DIFF
--- a/bin/fix-permissions
+++ b/bin/fix-permissions
@@ -15,7 +15,7 @@ if ! [ -e "$1" ] ; then
 fi
 
 find -L "$1" \! -gid 0 -exec chgrp 0 {} +
-find -L "$1" \! -perm /g+rw -exec chmod g+rw {} +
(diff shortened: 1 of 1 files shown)


jim-minter:
@hhorak assuming this is good to merge, please will you then rebuild and push all builder images affected by this?
@bparees fyi

bparees:
[test]

hhorak:
Yes, thanks for spotting this, I'll merge once we get tests results.

jim-minter:
please delay merging for a moment, I just want to confirm one more thing here

jim-minter:
ok, all good here, tests have passed, please merge if happy

bparees:
tests passed. merging.

PR #126, opened 2017-06-21
Fix warning messages when fix-permissions run on a symlink

omron93:
Change `fix-permissions` script to operate only with files which user running this script is owning. If `root` user is running the script all files are OK.

This fixes "Operation not permitted" errors from chgrp/chmod commands

Resolves: #112

### DIFF
--- a/bin/fix-permissions
+++ b/bin/fix-permissions
@@ -6,6 +6,8 @@ set +e
 # Fix permissions on the given directory or file to allow group read/write of
 # regular files and execute of directories.
 
+[ "$(id -u)" -ne 0 ] && CHECK_OWNER=" -uid $(id -u)"
+
 # If argument does not exist, script will still exit with 0,
 # but at least we'll see something went wrong in the log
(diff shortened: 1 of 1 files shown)


bin/fix-permissions
```diff
@@ -6,6 +6,8 @@ set +e
 # Fix permissions on the given directory or file to allow group read/write of
 # regular files and execute of directories.
 
+[ "$(id -u)" -ne 0 ] && CHECK_OWNER=" -uid $(id -u)"
```
pkubatrh: I don't think the double quotes in the test are needed. Otherwise lgtm

GrahamDumpleton:
Copying a version of this into ``/opt/app-root/bin`` from start of custom ``assemble`` wrapper, before calling original ``assemble``, in a repo processed through a Python S2I builder gets rids of the warnings. So first glance it appears to work.

pkubatrh:
lgtm. Merging, thanks.

PR #127, opened 2017-06-21
Don't show warning messages when fix-permissions run on a symlink

omron93:
Don't show error messages from fix-persmission script

(it tries to change permissions of all files accessible from given
directory - so it follow links and tries to change system files)

Resolves: #112


pkubatrh:
Would use `-f, --silent, --quiet` when calling chmod/chgrp instead of redirecting to `/dev/null` since that could also mask any potential `find` errors.

PR #128, opened 2017-06-28
Removing comments inside the LABEL command

pkubatrh:
OSBS seems to still have issues parsing the labels when there are comments included.


omron93:
LGTM, thanks.

PR #130, opened 2017-07-31
Move to newest common scripts

pkubatrh:
VERSION no longer used in make


omron93:
LGTM + tests are passing. So merging.

Thanks.

PR #131, opened 2017-08-20
Add fedora variants of the Dockerfiles

hhorak:


hhorak:
We also have Fedora Container Review: https://bugzilla.redhat.com/show_bug.cgi?id=1483351
We need s2i-core in Fedora so we can use s2i-core in Fedora variants of other images (like mongodb, nginx, ...)

omron93:
Fedora review done.

@hhorak Please update few missing things:

- run or usage label is missing (https://fedoraproject.org/wiki/Container:Guidelines#LABELS)
- help.1 or README.md is missing (https://fedoraproject.org/wiki/Container:Guidelines#Help_File)

As mentioned by @pkubatrh in https://github.com/sclorg/s2i-base-container/pull/132#issuecomment-323657039 - "s2i-core also needs the workaround applied for the image to be built successfully."

PR #134, opened 2017-10-06
Add rpm-file-permissions

pkubatrh:
As we are COPYing in some directories that are provided by the filesystem
rpm package we are unfortunately changing default permissions set on those
directories. This shell script uses information provided by an rpm query
to reset permissions of directories we are changing to their default values.

Fixes: rhbz#1481808

### DIFF
--- a/core/root/usr/bin/rpm-file-permissions
+++ b/core/root/usr/bin/rpm-file-permissions
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+CHECK_DIRS="/ /opt /usr /usr/bin /usr/lib /usr/lib64"
+
+rpm_format="[%{FILESTATES:fstate}  %7{FILEMODES:octal} %{FILENAMES:shescape}\n]"
+
+rpm -q --qf "$rpm_format" filesystem | while read line
+do
+    eval "set -- $line"
+
+    case $1 in
+        normal) ;;
+        *) continue ;;
+    esac
+
+    case " $CHECK_DIRS " in
+        *" $3 "*)
+            chmod "${2: -4}" "$3"
+            ;;
+    esac
+done
--- a/core/root/usr/bin/rpm-file-permissions
+++ b/core/root/usr/bin/rpm-file-permissions
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-CHECK_DIRS="/ /opt /usr /usr/bin /usr/lib /usr/lib64"
+CHECK_DIRS="/ /opt /etc /usr /usr/bin /usr/lib /usr/lib64"
 
 rpm_format="[%{FILESTATES:fstate}  %7{FILEMODES:octal} %{FILENAMES:shescape}\n]"
 
--- a/core/root/usr/bin/rpm-file-permissions
+++ b/core/root/usr/bin/rpm-file-permissions
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-CHECK_DIRS="/ /opt /etc /usr /usr/bin /usr/lib /usr/lib64"
+CHECK_DIRS="/ /opt /etc /usr /usr/bin /usr/lib /usr/lib64 /usr/share /usr/libexec"
 
 rpm_format="[%{FILESTATES:fstate}  %7{FILEMODES:octal} %{FILENAMES:shescape}\n]"
 
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -27,8 +27,6 @@ ENV \
     HOME=/opt/app-root/src \
     PATH=/opt/app-root/src/bin:/opt/app-root/bin:$PATH
 
-# Copy extra files to the image.
-COPY ./root/ /
 
 # When bash is started non-interactively, to run a shell script, for example it
 # looks for this variable and source the content of this file. This will enable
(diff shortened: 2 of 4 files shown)


core/root/usr/bin/rpm-file-permissions
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+CHECK_DIRS=${@-"/ /usr/bin /usr/lib /usr/lib64"}
+
+for dir in $CHECK_DIRS; do
+    echo "  -> Resetting permissions for $dir ..."
+    chmodcmd=$(rpm -q --qf "[%{FILESTATES:fstate}  %7{FILEMODES:octal} %{FILENAMES:shescape}\n]" filesystem|awk '/normal/{print "chmod", $2, $3}' | grep "'$dir'" | sed 's/chmod ./chmod /')
+    eval $chmodcmd
```
praiskup: Should be safe to use double-quotes here.

pkubatrh: Double-quotes added, thanks!

core/root/usr/bin/rpm-file-permissions
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+CHECK_DIRS=${@-"/ /usr/bin /usr/lib /usr/lib64"}
+
+for dir in $CHECK_DIRS; do
+    echo "  -> Resetting permissions for $dir ..."
+    chmodcmd=$(rpm -q --qf "[%{FILESTATES:fstate}  %7{FILEMODES:octal} %{FILENAMES:shescape}\n]" filesystem|awk '/normal/{print "chmod", $2, $3}' | grep "'$dir'" | sed 's/chmod ./chmod /')
```
praiskup: Sorry for not spotting it before, but some more nits:
- this runs the `rpm` for each dir in $CHECK_DIRS
- we might want to restore not only directories
- the sed chmod is weird, let's use awk/shell too

E.g.:
```bash
CHECK_DIRS="/ /usr/bin /usr/lib /usr/lib64"

rpm_format="[%{FILESTATES:fstate}  %7{FILEMODES:octal} %{FILENAMES:shescape}\n]"

rpm -q --qf "$rpm_format" filesystem | while read line
do
    eval "set -- $line"

    case $1 in
        normal) ;;
        *) continue ;;
    esac

    case " $CHECK_DIRS " in
        *" $3 "*)
            echo chmod "${2: -4}" "$3"
            ;;
    esac
done
```

praiskup: - The awk had an issue since the path could contain spaces ... and thus awk could operate on $4 and more..

pkubatrh:
Actually this is not working as expected:

```
make tag VERSIONS=core && docker run --rm -ti centos/s2i-core-centos7:latest bash -c "rpm -V filesystem"
...
.M.......    /
missing     /boot
.M.......    /opt
.M.......    /usr
```

So only `/usr/bin` had its permissions changed...

pkubatrh:
Fixed some of the issues I have encountered in previous comment (by resetting permissions last). Also moved some commands around in the Dockerfile to achieve better cache usage.

pkubatrh:
For some reason we still are unable to change permissions on `/` (even the rhel base image is affected). I guess this is due to some runtime checks of Docker itself but will look for more information.

pkubatrh:
>For some reason we still are unable to change permissions on /

Running `docker commit` on a container with permissions on `/` changed runs into the same issue. Docker does not seem to pick up that `/` has been changed and just uses the base image as is, without adding any additional layers on top.

However, this is only happening with some storage drivers - I can reproduce on `overlay2` and Pavel mentioned the same for `btrfs`. When `devicemapper` is used `chmod 0555 /` works as expected...

pkubatrh:
I created a BZ against rhel7 for the `/` issue over at: https://bugzilla.redhat.com/show_bug.cgi?id=1499755

praiskup:
Good job, LGTM

omron93:
So in every image using `s2i-core` base image where `root/` is added into `/`, `rpm-file-permissions` have to be run in each image to have right permissions to, right?

pkubatrh:
>So in every image using s2i-core base image where root/ is added into /, rpm-file-permissions have to be run in each image to have right permissions to, right?

Ultimately yes. So far it seems only `/usr/bin` is posing issues during certification.

pkubatrh:
Please allow for #136 to be merged in first as the s2i-base tests are actually not running against changes to s2i-core that are part of this PR.

praiskup:
[test]
And I'll play with this in https://github.com/sclorg/postgresql-container/pull/192
But I don't think we necessarily have to wait for the CI results.

pkubatrh:
Rebased. I want to wait at least for the centos pass on the newest state of this PR (since we are actually testing s2i-base built from s2i-core now).
I will test on rhel locally while this runs.

praiskup:
After several hours I canceled the build, seems like there's something wrong with Nodejs container build:
```
4:24:45 ---> Building your Node application from source
14:24:46 E1011 16:24:46.003688 05302 util.go:91] npm WARN package.json node-echo@0.0.1 No license field.
18:19:31 Build was aborted
18:19:31 Aborted by Pavel
```

pkubatrh:
I have tried running tests against both rhel and centos locally and they completed just fine

pkubatrh:
Will be merging this despite the failing test on rhel. The changes should not introduce any issues as far as I can tell and I have verified that the tests pass locally on both rhel and centos so it should be safe.

PR #136, opened 2017-10-10
hack: build and tag core first before attempting to build base

pkubatrh:
@praiskup PTAL

I wonder if this can be done in a nicer way without overriding the original rule...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,3 +7,10 @@ VERSIONS = core base
 .PHONY: $(shell test -f common/common.mk || echo >&2 'Please do "git submodule update --init" first.')
 
 include common/common.mk
+
+# HACK: We need to build core first and tag it right after, so that base picks it up in the same run
+# We cannot just depend on tag here since tag depends on build
+base: core
+core: core/root/help.1
+	VERSIONS="core" $(script_env) $(build)
+	VERSIONS="core" $(script_env) $(tag)


Makefile
```diff
@@ -7,3 +7,10 @@ VERSIONS = core base
 .PHONY: $(shell test -f common/common.mk || echo >&2 'Please do "git submodule update --init" first.')
 
 include common/common.mk
+
+# HACK: We need to build core first and tag it right after, so that base picks it up in the same run
+# We cannot just depend on tag here since tag depends on build
+base: core
```
praiskup: Hm, we don't have 'tag-core' (generated) target to depend on :( yet.

praiskup:
This is work-around which "knows too much" about the internals, and should work.  Long-term fix could be to have `tag-core` and `tag-base` targets (generated similarly to $(VERSIONS)) and we could depend on it.

praiskup:
LGTM

pkubatrh:
>Long-term fix could be to have tag-core and tag-base targets (generated similarly to $(VERSIONS)) and we could depend on it.

As this solution has its own problems and would additionally need changes in container-common-scripts, I will go with what is already in this PR (overriding default rules) and fix it properly in the future, once we find the right way to do it.

pkubatrh:
The failure in rhel CI is outside of the build so I think this is an issue with the CI and not the PR (similarly to what was encountered in https://github.com/sclorg/postgresql-container/pull/188) and should be safe to merge as any problems with the build process would likely come up on centos as well.

praiskup:
Agreed.

PR #139, opened 2017-11-03
Added APP_ROOT environment variable and replaced all occurences

ficap:


omron93:
@ficap Referencing variable declared in same `ENV` statement is not working.

First `ENV` has to be used to declare variable. And in next Dockerfile instructions you can use it.

ficap:
Good point, It's logical I forgot about it :)

pkubatrh:
lgtm thanks! Merging

PR #143, opened 2017-11-16
Define NODEJS_SCL environment variable

hhorak:
Define `$NODEJS_SCL` environment variable for NODEJS collection name, so depended images can work with it.


hhorak:
@pkubatrh I think it would be good to have this merged for the current respin, so we can start using the variable in the other containers.. could you, please, take a loot and merge if it seems fine?

omron93:
LGTM, thanks.

pkubatrh:
lgtm as well, thanks! Merging

PR #147, opened 2017-12-12
Change default shell to bash

omron93:
See https://docs.docker.com/engine/reference/builder/#shell

-  This allows to run scl command directly in Dockerfile `RUN` (no need to wrap it in scripts with `#!/bin/bash` shebang)
- This does not fix `docker exec`

Fixes: https://github.com/sclorg/s2i-python-container/issues/226

@hhorak @praiskup @pkubatrh Please take a look.

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -1,6 +1,8 @@
 # This image is the base image for all s2i configurable Docker images.
 FROM centos:7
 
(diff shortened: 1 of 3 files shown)


hhorak:
It looks good, let's try it and see whether it makes world great again :)

PR #149, opened 2017-12-16
Do not dissable all repos every-time

### LINKED ISSUE #148: Disabling all repositories in RHEL Dockerfiles is problematic
The cases where we need to enable different set of repositories that what we expect -- then disabling all repos is not possible. A usual trick to build RHEL dockerfiles on an unsubscribed machine is to add a different repo in the beginning of the Dockerfile. Since RHEL Dockerfiles currently disable all repos and enable just specific ones, it means the additional repo file won't be used.

A solution for this might be either of those (or both):
* disable only repos that are coming from subscription (those matching `rhel-*`)
* disable repos only when on subscribed machine (is a simple check like `[ -f /etc/pki/entitlement/*.pem ]` good enough?)

If we use both, it could look like:
```
RUN [ -f /etc/pki/entitlement/*.pem ] && yum-config-manager --disable rhel-*
```

hhorak:
It causes troubles when some other repositories are added by user or build system.
More discussion happens in #148.
Fixes #148.

### DIFF
--- a/core/root/usr/bin/prepare-yum-repositories
+++ b/core/root/usr/bin/prepare-yum-repositories
@@ -8,8 +8,8 @@ set -e
 # be used if we need to change this behaviour.
 # Once we realize there are real use cases for using those variables, we should
 # document them properly.
-DEFAULT_REPOS="rhel-7-server-rpms rhel-7-server-optional-rpms"
-SKIP_REPOS_ENABLE=false
+DEFAULT_REPOS=${DEFAULT_REPOS:-"rhel-7-server-rpms rhel-7-server-optional-rpms"}
+SKIP_REPOS_ENABLE=${SKIP_REPOS_ENABLE:-false}
 
 function is_subscribed() {
   for f in /run/secrets/etc-pki-entitlement/*.pem ; do
@@ -19,12 +19,14 @@ function is_subscribed() {
   return 1
 }
 
+# install yum-utils for yum-config-manager
 yum install -y yum-utils
+
 if is_subscribed ; then
   # Disable only repos that might come from subscribed host, because there
   # might be other repos provided by user or build system
-  echo "Disabling repositories rhel-\*"
-  yum-config-manager --disable rhel-\* &> /dev/null
+  echo "Running: yum-config-manager --disable rhel-\*"
+  yum-config-manager --disable rhel-\* &> /dev/null || :
 fi
 
 repos="${DEFAULT_REPOS:-}"
@@ -33,5 +35,6 @@ if [ $# -gt 0 ] ; then
 fi
 
 if [ ${SKIP_REPOS_ENABLE} = false -a -n "${repos}" ] ; then
-  yum-config-manager --enable "${repos}"
+  echo "Running: yum-config-manager --enable ${repos}"
+  yum-config-manager --enable ${repos}
 fi
--- a/core/root/usr/bin/prepare-yum-repositories
+++ b/core/root/usr/bin/prepare-yum-repositories
@@ -2,7 +2,7 @@
 
 # This script is used to prepare yum repositories, that are given as arguments.
 
-set -xe
+set -e
 
 function is_subscribed() {
   for f in /run/secrets/etc-pki-entitlement/*.pem ; do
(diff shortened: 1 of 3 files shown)


pkubatrh:
rhel7 test run fails because of

```
Installed: libselinux-2.5-12.el7.x86_64 (@rhel-7-server-htb-rpms)
```

hhorak:
It looks like this check does not work as I thought: `test -f '/etc/pki/entitlement/*.pem'`

pkubatrh:
>It looks like this check does not work as I thought: test -f '/etc/pki/entitlement/*.pem'

That is likely, since `test` only takes a single argument. How about using this instead?

```
ls /etc/pki/entitlement/*.pem 2>/dev/null
```

hhorak:
`ls` should work as well, although "the internet" suggests using `for`, which might work better in some corner cases, since it does not require stderr&stdout redirection to `/dev/null`.

pkubatrh:
[test]

hhorak:
We can have the list of repos in the script, since it is same in all containers (almost). Then, when beta needs to be used, we might set some argument for the build (taken from the builder environment)..

hhorak:
> We can have the list of repos in the script, since it is same in all containers (almost).

Actually it's not, I have a second thoughts about containers like devtoolset or rust/go/llvm from devtools channel and others, potentially even not from RHSCL, which would still be fine with s2i-core and s2i-base, but will definitely have different set of repositories. So, I'm actually staying with the original design and every container should specify the list of repos manually.

pkubatrh:
How about having the list of common repositories (rhel-7-server-rpms rhel-7-server-optional-rpms) included directly inside the new script and additional repositories (rhscl, devtools) to be enabled would be added as arguments to the script?

praiskup:
Seems like a nice idea to bake the list into the script.  Otherwise looks fine.

hhorak:
Ok, let's have the `rhel-7-server-rpms` and `rhel-7-server-optional-rpms` be the default set. The last commit also adds two intentionally undocumented variables that might be used for changing the behaviour of the script. They may be documented once we figure out that there are some real use cases for those variables.

core/root/usr/bin/prepare-yum-repositories
```diff
@@ -0,0 +1,37 @@
...
+  yum-config-manager --disable rhel-\* &> /dev/null
+fi
+
+repos="${DEFAULT_REPOS:-}"
+if [ $# -gt 0 ] ; then
+  repos="${repos:-} $@"
+fi
+
+if [ ${SKIP_REPOS_ENABLE} = false -a -n "${repos}" ] ; then
+  yum-config-manager --enable "${repos}"
```
praiskup: Slightly cleaner version (without the need to have `$repos` variable):
```bash
if ! ${SKIP_REPOS_ENABLE-false}; then
  yum-config-manager --enable $DEFAULT_REPOS "$@"
fi
```

core/root/usr/bin/prepare-yum-repositories
```diff
@@ -0,0 +1,37 @@
...
+
+# This script is used to prepare yum repositories, that are given as arguments.
+
+set -e
+
+# DEFAULT_REPOS and SKIP_REPOS_ENABLE are intentionally undocumented, but might
+# be used if we need to change this behaviour.
+# Once we realize there are real use cases for using those variables, we should
+# document them properly.
+DEFAULT_REPOS="rhel-7-server-rpms rhel-7-server-optional-rpms"
```
praiskup: This defines unconditionally, you can use:
```
: ${DEFAULT_REPOS-"rhel-7-server-rpms rhel-7-server-optional-rpms"}
```

core/root/usr/bin/prepare-yum-repositories
hhorak: Thanks, will use this.. :)

praiskup:
looks fine

pkubatrh:
[test]

pkubatrh:
lgtm as well. The test failures are because of a bug in varnish (https://github.com/sclorg/varnish-container/issues/31). Other images before varnish have been successfully tested so it is unlikely that this introduces any issues.

Merging.

Issue #150, opened 2017-12-18
docker run <scl container> no longer prints usage


jim-minter:
This is causing a raft of OpenShift tests to break.  I think https://github.com/sclorg/s2i-base-container/pull/147 is the problem commit.  Please revert #147, rebuild and repush all centos SCL images.
@bparees @coreydaley @omron93 @hhorak fyi

jim-minter:
Example failures: see https://ci.openshift.redhat.com/jenkins/view/All/job/test_branch_origin_extended_image_ecosystem/321/

bparees:
i don't know if #147 should be reverted but it would be good to get the images printing their usage by default again.

and thank you @jim-minter for tracking it down.

bparees:
@jim-minter @coreydaley in the meantime let's temporarily disable the usage tests in our bucket so we can get rid of the failure noise.

omron93:
@jim-minter  For example `docker run centos/python-36-centos7` prints usage normally. What are you testing?

pkubatrh:
Seems like this only affects some of the images. I can see this in `centos/perl-524-centos7` but not in `centos/python-36-centos7` as @omron93 points out

omron93:
My fault. I've pulled new image but used old one:-( Python image is affected as well...

omron93:
IMPOV it is bug in upstream docker

Reproducer:
```bash
$ cat Dockerfile
FROM centos:centos7
SHELL ["/usr/bin/bash", "-c"]
RUN echo "echo 'Default cmd!'" > /usr/bin/default-cmd && chmod a+x /usr/bin/default-cmd
ENV PREFIX=/usr/bin/
CMD ["/usr/bin/ls"]
$ cat Dockerfile.2 
FROM myimage
CMD $STI_SCRIPTS_PATH/default-cmd
$ docker build -t myimage -f Dockerfile . && docker build -t myimage2 -f Dockerfile.2 .
```

`docker inspect myimage2` contains in `Config` section:
```
"Cmd": [
                "/usr/bin/bash",
                "-c",
                "#(nop) "
            ],
```

So default CMD is noop.

omron93:
Upstream issue - https://github.com/moby/moby/issues/35832

I'll revert the change in s2i-core.

frenzymadness:
I've tested some S2I images I currently have locally available and all of them are showing the usage when invoked directly.

PR #151, opened 2017-12-19
Revert "Change default shell to bash"

omron93:
This reverts commit 744ed15d6abf3a6179cbb6fc66717ab051c6d2d9.

For more information see https://github.com/sclorg/s2i-base-container/issues/150


hhorak:
LGTM. RHEL tests fail from another reason, and since it's reverting to a state that worked before, let's merge it.

PR #153, opened 2018-02-06
Skip repository enable if there are additional .repo files

pkubatrh:
Related: https://github.com/sclorg/container-common-scripts/issues/25

### DIFF
--- a/core/root/usr/bin/prepare-yum-repositories
+++ b/core/root/usr/bin/prepare-yum-repositories
@@ -19,6 +19,9 @@ function is_subscribed() {
   return 1
 }
 
+# repo files might be mounted in (read-only), better not try any write ops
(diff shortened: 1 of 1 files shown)


praiskup:
The script is a bit of magic/fuzzy now (e.g. on subscribed machine, won't the disable `rhel-*` fail?) but it should do what we wanted in commons#25 -- let's assume that people either have subscribed host, or they provide the repo(s), not both.  I would also prefer other parse methods than `grep -v PATTERN`... but it LGTM anyways.

pkubatrh:
>The script is a bit of magic/fuzzy now (e.g. on subscribed machine, won't the disable rhel-* fail?)

Yes, I was aiming to tackle this one use case at a time, but I guess I might as well add the groundwork for this in this PR as well.

PR #154, opened 2018-02-07
Make option to control treatment of symbolic links in fix-permissions script adjustable

mcyprian:
Following symlinks in fix-permissions script can sometimes make troubles. Example of issues that can occur is [here](https://bugzilla.redhat.com/show_bug.cgi?id=1536129). Let's make it possible to pass the option to script as a command line argument.

### DIFF
--- a/core/root/usr/bin/fix-permissions
+++ b/core/root/usr/bin/fix-permissions
@@ -3,6 +3,8 @@
 # Allow this script to fail without failing a build
 set +e
 
+SYMLINK_OPT=${2:--L}
+
 # Fix permissions on the given directory or file to allow group read/write of
 # regular files and execute of directories.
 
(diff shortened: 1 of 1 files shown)


core/root/usr/bin/fix-permissions
```diff
@@ -3,6 +3,11 @@
 # Allow this script to fail without failing a build
 set +e
 
+SYMLINK_OPT=-L
+if [ $# -gt 1 ]; then
+  SYMLINK_OPT="$2"
+fi
```
pkubatrh: How about `SYMLINK_OPT=${2:--L}` instead of the whole block?

core/root/usr/bin/fix-permissions
```diff
@@ -16,10 +21,10 @@ if ! [ -e "$1" ] ; then
...
 fi
 
-find -L "$1" ${CHECK_OWNER} \! -gid 0 -exec chgrp 0 {} +
-find -L "$1" ${CHECK_OWNER} \! -perm -g+rw -exec chmod g+rw {} +
-find -L "$1" ${CHECK_OWNER} -perm /u+x -a \! -perm /g+x -exec chmod g+x {} +
-find -L "$1" ${CHECK_OWNER} -type d \! -perm /g+x -exec chmod g+x {} +
+find "$SYMLINK_OPT" "$1" ${CHECK_OWNER} \! -gid 0 -exec chgrp 0 {} +
+find "$SYMLINK_OPT" "$1" ${CHECK_OWNER} \! -perm -g+rw -exec chmod g+rw {} +
+find "$SYMLINK_OPT" "$1" ${CHECK_OWNER} -perm /u+x -a \! -perm /g+x -exec chmod g+x {} +
+find "$SYMLINK_OPT" "$1" ${CHECK_OWNER} -type d \! -perm /g+x -exec chmod g+x {} +
```
praiskup: IMO you don't want to have doublequotes here... Because then find always has (even though empty) option passed through execve().

core/root/usr/bin/fix-permissions
praiskup: Another alternative: `SYMLINK_OPT=${SKIP_SYMLINKS--L}`.

praiskup: ... that would have to be explicitly specified in dockerfiles, which might be worth.  Dunno.  What @pkubatrh suggests is good, too.

pkubatrh: >Another alternative: SYMLINK_OPT=${SKIP_SYMLINKS--L}.

If it got defined as an `ENV` it would get baked into the image and used even for s2i-builds. If it was just used for the fix-permissions call then I do not see much difference apart from being a bit more verbose (`SKIP_SYMLINKS=1` vs `-P`)

pkubatrh:
[test]

praiskup:
lgtm

mcyprian:
Do you have any additional comments, questions? Can I consider this to be approved?

pkubatrh:
Yeah sorry, was waiting for the CI to finish before merging (takes a few hours).

Issue #158, opened 2018-03-06
optimize the prepare-yum-repositories script


praiskup:
Seems like in our CI, the script costs about 3 minutes per each container version; so
for sclorg/postgresql-container/pull/242 (4 versions) it takes about 12 minutes only
on disabling/enabling repositories.

praiskup:
Maybe this is about optimizing the VM images, dunno.  Just wanted to track this somewhere.

pkubatrh:
Limiting the disable to only enabled repositories might help a bit.

Something along the lines of (for my box at least):
`yum-config-manager --disable $(yum repolist -v | grep Repo-id | awk '{ print $3 }' | awk -F / '{ print $1 }' | xargs)`

praiskup:
The speedup is just awesome, so I would go for the optimization (even though
it looks a bit over-engineered).

PR #159, opened 2018-03-06
Optimize 'prepare-yum-repositories' script

praiskup:
.. so we disable only enabled repositories, without fnmatch()
pattern so the yum-config-manager doesn't have to(?) iterate
through all repositories.

Per @pkubat's suggestion.

Fixes: #158

### DIFF
--- a/core/root/usr/bin/prepare-yum-repositories
+++ b/core/root/usr/bin/prepare-yum-repositories
@@ -29,8 +29,19 @@ yum install -y yum-utils
 if [ "$SKIP_REPOS_DISABLE" = false ] && is_subscribed; then
   # Disable only repos that might come from subscribed host, because there
   # might be other repos provided by user or build system
-  echo "Running: yum-config-manager --disable rhel-\*"
-  yum-config-manager --disable rhel-\* &> /dev/null || :
+
+  disable_repos=
+  # Lines look like: "Repo-id  :  dist-tag-override/x86_64"
+  while IFS=' /' read -r _ _ repo_id _; do
+      case $repo_id in rhel-*)
+        disable_repos+=" $repo_id" ;;
+      esac
+  done <<<"$(yum repolist -v 2>/dev/null | grep Repo-id)"
+
+  if test -n "$disable_repos"; then
+    echo "Disabling repositories:$disable_repos"
+    yum-config-manager --disable $disable_repos &> /dev/null
+  fi
 fi
 
 if [ ${SKIP_REPOS_ENABLE} = false ] && [ -n "${DEFAULT_REPOS}" -o $# -gt 0 ] ; then


core/root/usr/bin/prepare-yum-repositories
```diff
@@ -29,8 +29,23 @@ yum install -y yum-utils
...
+              disable_repos+=" $repo_id"
+              ;;
+      esac
+  done <<<"$(yum repolist -v 2>/dev/null | grep Repo-id)"
+
+  if test -n "$disable_repos"; then
+    echo "Disabling repositories:$disable_repos"
+    if ! yum-config-manager --disable $disable_repos &> /dev/null; then
+        # Shoudln't we rather fail the build?
+        echo "... diabling failed."
```
praiskup: There used to be `|| :`, but I don't know why, @hhorak?

praiskup:
FTR, after this fix the script took 13s: `16:03:25 => 16:03:48`, per: https://coreservices-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/rhscl-images-s2i-rh/164/consoleText

praiskup:
And FTR when I reverted `16:13:47 => 16:16:46, almost 3 minutes, per: https://coreservices-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/rhscl-images-s2i-rh/165/console

pkubatrh:
[test]

core/root/usr/bin/prepare-yum-repositories
```diff
@@ -29,8 +29,22 @@ yum install -y yum-utils
...
+      case $repo_id in rhel-*)
+        disable_repos+=" $repo_id" ;;
+      esac
+  done <<<"$(yum repolist -v 2>/dev/null | grep Repo-id)"
+
+  if test -n "$disable_repos"; then
+    echo "Disabling repositories:$disable_repos"
+    if ! yum-config-manager --disable $disable_repos &> /dev/null; then
+        # Shoudln't we rather fail the build?
+        echo "... but 'yum-config-manager --disable' FAILED"
```
praiskup: Ping on this, I used `if ...; then` condition to ignore non-zero exit status, before there
was `yum-config-manager ... || :`.  Is this needed?

pkubatrh: I dont think it is. When the host is not subscribed `yum-config-manager` still exits with 0. Same in the case when the repositories it tries to disable do not exist. Is there any other case we need to be care about?

@hhorak ?

praiskup: ping

hhorak: My intention with `|| :` was to avoid the case when there is no repository called 'rhel-*', which I read is not the case, so feel free to remove it.

praiskup:
Please have a look.

pkubatrh:
lgtm

omron93:
LGTM, thanks. No objections, so merging.

PR #162, opened 2018-03-27
Only check if redhat.repo exists in prepare-yum-repositories

pkubatrh:
Also run set -x to make sure we know what is happening during build

### DIFF
--- a/core/root/usr/bin/prepare-yum-repositories
+++ b/core/root/usr/bin/prepare-yum-repositories
@@ -2,7 +2,7 @@
 
 # This script is used to prepare yum repositories, that are given as arguments.
 
-set -e
+set -ex
 
 # DEFAULT_REPOS and SKIP_REPOS_{ENABLE,DISABLE} are intentionally undocumented,
 # but might be used if we need to change this behaviour.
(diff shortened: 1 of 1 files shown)


core/root/usr/bin/prepare-yum-repositories
```diff
@@ -2,7 +2,7 @@
 
 # This script is used to prepare yum repositories, that are given as arguments.
 
-set -e
+set -ex
```
praiskup: Do you really want to have the `-x` here?  If so, we pobably don't need to echos below.

praiskup:
TBH, the logic behind ENABLE/DISABLE is really hard to understand.  I always keep parsing that code over-again...  before LGTM.  So except for the `-x` nit, LGTM.

core/root/usr/bin/prepare-yum-repositories
pkubatrh: I have added that on purpose so that we can see what is happening during the build in detail. The script is important enough (and not that long) so that it should not hurt to have it more verbose.

I have however missed that there are echo commands present, will remove those. Thanks for the tip

pkubatrh:
>TBH, the logic behind ENABLE/DISABLE is really hard to understand.

Can you be more specific? I can add some comments where needed in order to make it easier to understand.

praiskup:
> # DEFAULT_REPOS and SKIP_REPOS_{ENABLE,DISABLE} are intentionally undocumented,

reading now ^^^, so it is actually intentionally undocumented.  I mean, I just don't see the motivation one of those variables;  aren't there just misleading?

praiskup:
LGTM

omron93:
LGTM too. Thanks.

pkubatrh:
> I mean, I just don't see the motivation to use one of those variables; aren't they just misleading?

iirc this was left undocumented mostly because we do not yet have a way to easily set these variables other than through changing the Dockerfile.

In any case I think this is out of the scope of this PR so I will merge it.

pkubatrh:
Thanks for the review!

PR #165, opened 2018-04-26
prepare-yum-repositories: cooperate with make CUSTOM_REPO=...

praiskup:
@caringi , @pkubatrh, that's what I talked about yesterday...  It detects
whether CUSTOM_REPO=/an/repo/file.repo or CUSTOM_REPO=/dir/repo/
was passed during 'make build'.

It's not nice, but it is neither the first hack in that file.  Comments welcome!

### DIFF
--- a/core/root/usr/bin/prepare-yum-repositories
+++ b/core/root/usr/bin/prepare-yum-repositories
@@ -1,6 +1,15 @@
 #!/bin/bash
 
 # This script is used to prepare yum repositories, that are given as arguments.
+# It is no-op if user also mounts the repo file(s) into the container during
+# image build.  This can be done by one of those commands:
+#
+#   docker build -v /some/repo/file:/etc/yum.repos.d/sclorg_custom.repo
+#   docker build -v /some/repo/directory:/etc/yum.repos.d
+#   make CUSTOM_REPO=/some/repo/file/or/directory
+#
+# The last one works for projects where we have Makefile with the
+# container-common-scripts support.
 
 set -ex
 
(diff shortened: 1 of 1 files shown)


omron93:
Looks good, thanks.

pkubatrh:
Looks good, I do not think we need the line below with this change:
```
[ -f /etc/yum.repos.d/redhat.repo ] || SKIP_REPOS_ENABLE=true
```

We should also document that the user has to provide a specific name for his repository if he wants to use the feature (outside CUSTOM_REPO which does it for him). Do we even have the documentation for yum-prepare-repositories somewhere?

praiskup:
On Monday, April 30, 2018 8:49:00 AM CEST Petr Kubat wrote:
> Looks good, I do not think we need the line below with this change:
> ```
> [ -f /etc/yum.repos.d/redhat.repo ] || SKIP_REPOS_ENABLE=true
> ```

So I dropped it.  Thanks.

> We should also document that the user has to provide a specific name for his
> repository if he wants to use the feature (outside CUSTOM_REPO which does it
> for him).

Documented.

> Do we even have the documentation for yum-prepare-repositories somewhere?

I don't think so, except for the self-documentation in the file itself.
That would be separate issue, though.

pkubatrh:
>I don't think so, except for the self-documentation in the file itself. That would be separate issue, though.

Definitely. I was just wondering where would be the best place to make the change but if we do not have anything yet it should suffice to have it in-code.

pkubatrh:
Fedora CI fails when trying to pull s2i-core - seems like it is not available in the registry (but should be). Not related to this PR so ignoring.

pkubatrh:
Works good with both `CUSTOM_REPO` use cases so lgtm, merging.

PR #168, opened 2018-05-17
Do not use scl_enable in Fedora Dockerfile

pkubatrh:


praiskup:
Please explain the reasons for candidate registry in the commit message, or maybe even in the Dockerfile, otherwise looks fine.

pkubatrh:
I used candidate registry because the build failed during pull so I thought s2i-core has not yet been built, but it seems to be working now... Will just default to regular registry instead.

praiskup:
Looks good to me, but centos7 CI is failing [test].

pkubatrh:
Seems like there are some dependency issues for varnish5 using candidate repository. Its rpms are however not released yet in regular repositories so not sure if we can work around that in any way.

pkubatrh:
@hhorak Any idea when the rhscl 3.1 rpms will get rebuilt in centos?

omron93:
[test]
[-varnish] - Skip varnish testing. This PR changes only Fedora images, so varnish isn't broken by this change.

omron93:
I wanted to add ability to skip testing of some dependent images in PR -- see https://github.com/sclorg/rhscl-container-ci/pull/18
(by dependent image I mean image which uses tested image as a base image - for example in case of this PR it's all s2i based images)
But I've realized that testing of dependent images are also run before pushing image to Docker Hub. Does we want this?
I can imagine work-flow:
1. testing of PR shows which dependent images fails to build after PR is merged and image built (it could be caused change in the PR or by something totally different like this rpm dependency problem)
2. before images is built only tests for image itself are run
3. then builds for all dependent images are triggered. So some images may fail to build (indicated in PR testing). But this won't block other images.

@pkubatrh @praiskup @hhorak Ideas?

pkubatrh:
I would slightly modify the workflow - instead of just running the tests for the base image lets also run the tests for layered images, ignoring the failures and creating issues for images that failed their test runs (unless such an issue is already created and not closed). Would that be possible?

omron93:
> I would slightly modify the workflow - instead of just running the tests for the base image lets also run the tests for layered images, ignoring the failures and creating issues for images that failed their test runs (unless such an issue is already created and not closed). Would that be possible?

IMHO it would be possible.

But I think it is not necessary to run tests for layered images during build if the failure won't be blocking. Because in case of nonblocking tests, build of the failing image will be run anyway. So if we want an issue, we can create it for every failed build. Do we want issues for failed build? Or jenkins email notification for maintainers?

pkubatrh:
>But I think it is not necessary to run tests for layered images during build if the failure won't be blocking. 

True, likely better to just create an issue for every failed build (which would be the case with varnish right now).

> Do we want issues for failed build? Or jenkins email notification for maintainers?

I guess this would depend on how much work would be needed to make issue creating work. Imo issues are more visible than email notifications so I would rather go with that option.

hhorak:
@pkubatrh, @khardix is working on them, I expect them every day. @khardix, any ETA?

khardix:
@hhorak @pkubatrh ETA is end of this week (hopefully); However, `rh-varnish5` currently failing its test suite and it is under investigation: https://ci.centos.org/view/SCLo-pkgs-cbs/job/SCLo-pkg-rh-varnish5-rh-C7-candidate-x86_64/34

pkubatrh:
>It seems that varnish dependency problem is solved.

Great! Will merge then.

pkubatrh:
@hhorak How soon can we get this change into the fedora candidate registry at least?

PR #170, opened 2018-05-23
Use nodejs v8 for having npm

hhorak:
There are more reasons why we benefit from switching to latest version -- at least later EOL, and availability on more arches than x86_64 only.


omron93:
Tests are passing and changes look good.
@hhorak Should this be tested somehow specially?

hhorak:
I just wanted to re-check that all s2i-based containers already use the env. variable and not hard-code the nodejs6.. so, please, don't merge for now, I can merge once I confirm the other images status.

hhorak:
Just checked, all should be fine.

hhorak:
[test]

PR #172, opened 2018-07-17
Install just groff-base, which is enough to have groff binary

hhorak:
SSIA


praiskup:
Looks OK, thanks.

pkubatrh:
Yes, makes sense.

The test suite is currently failing on golang tests - I guess a new version got released?

hhorak:
Merging despite the test suite is not entirely cool, but it shouldn't be caused by this issue.

PR #173, opened 2018-08-23
Move to F28 and install mariadb-connector-c-devel instead of mariadb-devel

hhorak:
ssia


omron93:
Thanks for the fix.

Tests failure isn't caused by this change. Centos7 and RHEL7 CI is failing because of failing golang-container testsuite (golang were updated to newer version and tests expects older one). So merging.

PR #177, opened 2018-11-27
Add rhel8 version for core and base

pkubatrh:


praiskup:
Can we setup CI for rhel8?

core/root/usr/bin/cgroup-limits
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/libexec/platform-python
```
praiskup: Is this expected?

praiskup: Aha, I see:
`[ -e /usr/libexec/platform-python ] || ln -s /usr/libexec/system-python /usr/libexec/platform-python`

pkubatrh:
@omron93 @hhorak?

I expect one would have to resolve the issue of not having the rhel8 based images in any public registry before we can do that however.

core/root/usr/bin/cgroup-limits
pkubatrh: Yes, this should only be needed for Fedora, as both rhel7 and rhel8 have `/usr/libexec/platform-python`

praiskup:
I have troubles to build the image, but the change looks good to me.  Working CI is not prerequisite.

core/Dockerfile.rhel8
```diff
@@ -0,0 +1,83 @@
...
+ENV BASH_ENV=${APP_ROOT}/etc/scl_enable \
+    ENV=${APP_ROOT}/etc/scl_enable \
+    PROMPT_COMMAND=". ${APP_ROOT}/etc/scl_enable"
+
+# Copy just prepare-yum-repositories that is needed for packages install step,
+# other files might be added later so changing them does not cause packages
+# to be installed again, which takes long time
+
+# TODO: Everything is in AppStream/BaseOS/Buildroot, do we still need this?
+# COPY ./root/usr/bin/prepare-yum-repositories /usr/bin/prepare-yum-repositories
```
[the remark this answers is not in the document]
pkubatrh: I will try looking into it. Probably no other way than installing the beta and trying to get subscriptions set up properly somehow. But the expectation of having the basic repositories (baseos and appstream) enabled by default seems sane to me for the time being.

core/Dockerfile.rhel8
```diff
@@ -0,0 +1,83 @@
...
+# This is the list of basic dependencies that all language container image can
+# consume.
+# Also setup the 'openshift' user that is used for the build execution and for the
+# application runtime execution.
+# TODO: Use better UID and GID values
+
+RUN INSTALL_PKGS="bsdtar \
+  findutils \
+  groff-base \
+  glibc-locale-source \
```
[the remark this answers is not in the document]
pkubatrh: `glic-locale-source, glibc-all-langpacks` were added as they were previously present in the rhel base image. Not sure about `xz`, likely added for the same reason but would have to double check.

core/Dockerfile.rhel8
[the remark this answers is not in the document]
pkubatrh: I managed to confirm that on a subscribed RHEL8 host both BaseOS and Appstream repositories are available and enabled by default.

omron93:
CI failures are caused by postgresql and depend on sclorg/postgresql-container/pull/303

praiskup:
[test]

core/Dockerfile.rhel8
```diff
@@ -0,0 +1,83 @@
...
+    HOME=/opt/app-root/src \
+    PATH=/opt/app-root/src/bin:/opt/app-root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+
+# When bash is started non-interactively, to run a shell script, for example it
+# looks for this variable and source the content of this file. This will enable
+# the SCL for all scripts without need to do 'scl enable'.
+# NOTE: The scl_enable script is "empty" most of the time. It is only present for images
+# that actually need to use scls in rhel8 (like devtools) or need to prepare their
+# environment before anything is run inside (python virtualenv).
+ENV BASH_ENV=${APP_ROOT}/etc/scl_enable \
```
omron93: This is going to cause problems for images that provide both rhel7 and rhel8 images! (`./root/opt/app-root/etc/scl_enable` is present for rhel7 image, but is going to be installed also into rhel8 image and fails there)

pkubatrh: `scl_enable` will get copied in only for images that actually use it (devtools), the rest will be using the s2i-core version of the script which is basically empty. I do not see how it would cause failures in images that have both rhel7 and rhel8 versions.

pkubatrh: Ah right, because we are copying in the whole `root` directory, which would mean `scl_enable` scripts as well...

omron93: I see two options here:
- go the same way as Fedora images: don't declare this in ENV in s2i-core and only declare them in images that needs it (for example https://github.com/sclorg/s2i-python-container/blob/master/3.6/Dockerfile.fedora#L65)
- change scl_enable in s2i-core to enable all collections defined in some variable - lets say ` ENABLED_COLLECTIONS`. And change all existing images not to add own scl_enable (except few images that do additional stuff there) but to define/add to `ENABLED_COLLECTIONS`
To note: some database images already declare `ENABLED_COLLECTIONS`, php uses `SCL_ENABLED`.

pkubatrh: >go the same way as Fedora images: don't declare this in ENV in s2i-core and only declare them in images that needs it

Yes this would work, however we would have to touch most of the Dockerfiles to make sure every Dockerfile is declaring ENVs properly. Also a lot of pain if we ever want to change it again in the future. Also in this case the scl_enable scripts would still be copied into the image.

>change scl_enable in s2i-core to enable all collections defined in some variable - lets say ENABLED_COLLECTIONS. And change all existing images not to add own scl_enable 

This looks a bit better - we would still have to touch most of the images to remove the scl_enable commands as well but at least this would not introduce more code to maintain.

Another option I can think of is to override the scl_enable scripts in the rhel8 Dockerfiles.

pkubatrh: Ideally we should resolve this by having better source generation, eg. having different directory trees for each version of the image.

pkubatrh:
After discussing this internally we have decided to just drop the scl_enable hack in the rhel8 Dockerfile. Afaik only the rhel8 versions of python images actually use the hack for something (virtualenv activation) so those two images can include it in their Dockerfiles.

base/Dockerfile.rhel8
```diff
@@ -0,0 +1,49 @@
...
+  gd-devel \
+  gdb \
+  git \
+  libcurl-devel \
+  libpq-devel \
+  libxml2-devel \
+  libxslt-devel \
+  lsof \
+  make \
+  mariadb-devel \
```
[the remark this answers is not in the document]
praiskup: I don't think we need this, unless we want to be able to build server-side modules;
i.e. we also don't install `postgresql-server-devel` package.

core/Dockerfile.rhel8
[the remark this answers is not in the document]
praiskup: So what about to drop the TODO item and commented COPY instruction right away?

praiskup:
The two minor things, otherwise it looks fine to me.

pkubatrh:
Thanks for the reviews everyone! Will wait until tests are done and will merge if green.

core/root/usr/bin/cgroup-limits
hhorak: I'm afraid there is no `/usr/libexec/platform-python` in RHEL 7 and CentOS 7.

PR #179, opened 2019-01-22
Start depending on Nodejs v10 for npm

hhorak:
Because Nodejs will EOL at some point and we have nodejs v10 already.


hhorak:
Tests pass, let's merge, so it gets to the next respin.

Issue #182, opened 2019-02-13
Define environment variables with OS version


hhorak:
Some scripts currently use something like this:
```
if head "/etc/redhat-release" | grep -q "^Red Hat Enterprise Linux release 8"; then
```
I think it would be easier to have some variable like `PLATFORM_ID="platform:f28"` (defined in `/etc/os-release`) to be defined in s2i-core image. The condition above would look like this then:

```
if [ "${PLATFORM_ID}" == "platform:el8" ] ; then
```

PR #183, opened 2019-03-06
Add platform identifier to Dockerfile as ENV

pkubatrh:
Resolves: #182

### DIFF
--- a/core/Dockerfile.fedora
+++ b/core/Dockerfile.fedora
@@ -29,7 +29,8 @@ ENV \
     APP_ROOT=/opt/app-root \
     # The $HOME is not set by default, but some applications needs this variable
     HOME=/opt/app-root/src \
-    PATH=/opt/app-root/src/bin:/opt/app-root/bin:$PATH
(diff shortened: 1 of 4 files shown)


core/Dockerfile.fedora
```diff
@@ -29,7 +29,8 @@ ENV \
     APP_ROOT=/opt/app-root \
     # The $HOME is not set by default, but some applications needs this variable
     HOME=/opt/app-root/src \
-    PATH=/opt/app-root/src/bin:/opt/app-root/bin:$PATH
+    PATH=/opt/app-root/src/bin:/opt/app-root/bin:$PATH \
+    PLATFORM="fedora"
```
omron93: `$FGC` could be used to get also the version of Fedora.

pkubatrh: I left the fedora string unversioned on purpose since we do not do anything specific to a single fedora release (afaik) and if it is really needed the `$FGC` variable can be used once we know we are running a fedora container.

PR #184, opened 2019-03-25
Add  s2i-core-ubi7 Dockerfiles

sub-mod:
The dev preview images are available to public `registry.access.redhat.com/ubi8-dev-preview/ubi:latest`
It would be nice if this scl images are available. helps to showcase s2i app examples with this image.

### DIFF
--- a/base/Dockerfile.rhel7
+++ b/base/Dockerfile.rhel7
@@ -15,7 +15,8 @@ LABEL summary="$SUMMARY" \
       io.k8s.display-name="s2i base" \
       com.redhat.component="s2i-base-container" \
       name="rhscl/s2i-base-rhel7" \
-      version="1"
+      version="1" \
+      com.redhat.license_terms="https://www.redhat.com/en/about/red-hat-end-user-license-agreements#UBI"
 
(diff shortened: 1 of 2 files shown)


pkubatrh:
Hi @sub-mod , thanks for the PR.

I am ok with adding a ubi8 version of the images but it would make more sense to me to just base the existing rhel8 Dockerfiles on the dev preview rather than introducing a new OS target. I expect only changing the `FROM` in `Dockerfile.rhel8` should be enough to make the image build. That would also make the PR to common scripts redundant.

Regarding the ubi7 base image, there are some issues with licensing that will need to be taken care of before we are able to build the rhscl images on top of an ubi7 base image so please drop these changes for now. If you want to track the ubi enablement for rhscl images feel free to open up a separate issue for it.

hhorak:
I believe we'll only maintain one copy of Dockerfiles for each, ubi7 and ubi8, once the images move to UBI.

pkubatrh:
Added a link to the ubi7 EULA to each image's Dockerfile.

[test]

pkubatrh:
Tests are green so merging, thanks!

PR #185, opened 2019-03-25
Add s2i-core-ubi8 image

sub-mod:
The dev preview images are available to public registry.access.redhat.com/ubi8-dev-preview/ubi:latest
It would be nice if this scl images are available. helps to showcase s2i app examples with this image.

This PR is based on feedback from https://github.com/sclorg/s2i-base-container/pull/184#issuecomment-476081284
will wait for license issue to clear up to add ubi7 image.


pkubatrh:
Added a link to the ubi7 EULA to each image's Dockerfile. Also moved the images to the `ubi8/` namespace.

However, we should wait until there is a working rhel8 CI before merging this, just in case.

PR #187, opened 2019-05-14
move rhel7-based s2i-core to ubi7

pkubatrh:


pkubatrh:
Closing in favour of #184

Issue #190, opened 2019-07-12
yum clean should work with all repos


hhorak:
There is an issue when something non-standard happens to RPM repos in the image, which has a consequence of not cleaning all repo cache properly and results in too large image:
https://bugzilla.redhat.com/show_bug.cgi?id=1729172

We should clean all repositories similarly as it is done in python image (`yum -y clean all --enablerepo='*'`) and possibly also include a check that `/var/cache/yum` or even `/var` is not unexpectedly big. This all should be done in all image.

PR #192, opened 2019-07-18
Clean yum cache for every available repository

pkubatrh:


pkubatrh:
Fixes: #190

pkubatrh:
Seems like the rhel7 tests are failing because of not being able to contact the cdn. Does not look like this is caused by this change though, as older PRs (eg. https://github.com/sclorg/devtoolset-container/pull/40) are failing as well.

hhorak:
[test]

PR #193, opened 2019-09-11
clean yum cache in rhel8 based s2i-core

pkubatrh:
use yum instead of dnf as yum is now installed in the base image by default

### DIFF
--- a/core/Dockerfile.rhel8
+++ b/core/Dockerfile.rhel8
@@ -52,9 +52,9 @@ RUN INSTALL_PKGS="bsdtar \
   yum" && \
   mkdir -p ${HOME}/.pki/nssdb && \
   chown -R 1001:0 ${HOME}/.pki && \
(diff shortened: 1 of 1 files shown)


pkubatrh:
rhel8 failure is from trying to build centos7 based cassandra for some reason, will investigate

pkubatrh:
rhel7 failure is from explicit shadow-utils install and rpm verification scheck even though it comes from base image. Removing in images where it is causing issues in separate PRs.

pkubatrh:
[test]

pkubatrh:
centos7 failure should be fixed by https://github.com/sclorg/mysql-container/commit/0ad545c7e1e8b4eccf5753b24f11f84a23ac1f90 but it seems the CI did not re-trigger

phracek:
LGTM. Thanks. I think it can be merged even w/o CI tests.

pkubatrh:
CI tests green now

PR #195, opened 2019-10-31
Add missing packages redhat-rpm-config and glibc-langpack-en

phracek:
Add `redhat-rpm-config` package into `s2i-base` image
and replace `glibc-langpack-en` package instead `glibc-all-langpacks`
in `s2i-core` image

Reference to s2i-python-container issue: https://github.com/sclorg/s2i-python-container/pull/346

### DIFF
--- a/base/Dockerfile.rhel8
+++ b/base/Dockerfile.rhel8
@@ -39,6 +39,7 @@ RUN yum -y module enable nodejs:$NODEJS_VER && \
   patch \
   procps-ng \
   npm \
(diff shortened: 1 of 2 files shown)


frenzymadness:
This change looks good to me and the test failure on Centos7 doesn't seem to be related to it. The failure is caused by non-working repository `cbs.centos.org_repos_sclo7-rh-postgresql10-rh-candidate_x86_64_os_`.

PR #198, opened 2020-01-13
Fix symlink from /usr/libexec/system-python to /usr/bin/python3

phracek:
Link /usr/libexec/platform-python is fixed. Now it references to /usr/bin/python3.

Required: s2i-ruby-container#248


pkubatrh:
Failure caused by a bug in ruby test suite. Not connected to this PR, so merging.

PR #199, opened 2020-02-06
move core and base to fc31

pkubatrh:
install glibc-langpack-en as it was removed from fedora base image
use libpq-devel instead of postgresql-devel


pkubatrh:
@phracek PTAL

pkubatrh:
both rhel7 and centos7 failing because of https://github.com/sclorg/s2i-ruby-container/pull/248

pkubatrh:
Not touching rhel7/centos7 at all so merging without waiting for ruby.

pkubatrh:
Thanks for review!

PR #202, opened 2020-06-22
Cgroups v2 support for cgroup-limits

### LINKED ISSUE #201: cgroup-limits do not work with kernel 5.6 (cgroups v2)
On Fedora 32 with kernel kernel-5.6.12-300.fc32.x86_64:
```
host#> podman run -ti --rm -m 128M --cpuset-cpus=1 ubi8 bash
cont#> curl https://raw.githubusercontent.com/sclorg/s2i-base-container/master/core/root/usr/bin/cgroup-limits >/usr/bin/cgroup-limits
cont#> chmod a+x /usr/bin/cgroup-limits
cont#> /usr/bin/cgroup-limits
Warning: Can't detect memory limit from cgroups
Warning: Can't detect number of CPU cores from cgroups
MAX_MEMORY_LIMIT_IN_BYTES=9223372036854775807
```

Digging why, it's probably because cgroups v2 is used now for this newer kernel
```
cont#> cat /sys/fs/cgroup/cpuset/cpuset.cpus
cat: /sys/fs/cgroup/cpuset/cpuset.cpus: No such file or directory
cont#> cat /sys/fs/cgroup/memory/memory.limit_in_bytes
cat: /sys/fs/cgroup/memory/memory.limit_in_bytes: No such file or directory
```

So what we probably need to do is to adopt the v2 structure:
```
cont#> cat /sys/fs/cgroup/memory.max
134217728
cont#> # cat /sys/fs/cgroup/cpuset.cpus
1
```

frenzymadness:
Fixes #201

### DIFF
--- a/core/root/usr/bin/cgroup-limits
+++ b/core/root/usr/bin/cgroup-limits
@@ -19,10 +19,12 @@ Variables currently supported:
         Maximum amount of user memory in bytes. If this value is set
         to the same value as MAX_MEMORY_LIMIT_IN_BYTES, it means that
         there is no limit set. The value is taken from
-        /sys/fs/cgroup/memory/memory.limit_in_bytes
+        /sys/fs/cgroup/memory/memory.limit_in_bytes for cgroups v1
+        and from /sys/fs/cgroup/memory.max for cgroups v2
     NUMBER_OF_CORES
         Number of detected CPU cores that can be used. This value is
-        calculated from /sys/fs/cgroup/cpuset/cpuset.cpus
+        calculated from /sys/fs/cgroup/cpuset/cpuset.cpus for cgroups v1
+        and from /sys/fs/cgroup/cpuset.cpus.effective for cgroups v2
     NO_MEMORY_LIMIT
         Set to "true" if MEMORY_LIMIT_IN_BYTES is so high that the caller
         can act as if no memory limit was set. Undefined otherwise.
@@ -49,6 +51,8 @@ def get_memory_limit():
     # If first file does not exist, try cgroups v2 file
     limit = limit or _read_file('/sys/fs/cgroup/memory.max')
     if limit is None or not limit.isdigit():
+        if limit == 'max':
+            return 9223372036854775807
         print("Warning: Can't detect memory limit from cgroups",
               file=sys.stderr)
         return None
@@ -64,8 +68,10 @@ def get_number_of_cores():
 
     line = _read_file('/sys/fs/cgroup/cpuset/cpuset.cpus')
     # If first file does not exist, try cgroups v2 file
-    line = line or _read_file('/sys/fs/cgroup/cpuset.cpus')
+    line = line or _read_file('/sys/fs/cgroup/cpuset.cpus.effective')
     if line is None:
+        # None of the files above exists when running podman as non-root,
+        # so in that case, this warning is printed every-time
         print("Warning: Can't detect number of CPU cores from cgroups",
               file=sys.stderr)
         return None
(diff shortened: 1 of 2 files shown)


core/root/usr/bin/cgroup-limits
```diff
@@ -61,6 +63,8 @@ def get_number_of_cores():
     core_count = 0
 
     line = _read_file('/sys/fs/cgroup/cpuset/cpuset.cpus')
+    # If first file does not exist, try cgroups v2 file
+    line = line or _read_file('/sys/fs/cgroup/cpuset.cpus')
```
hhorak, line 67: This currently fails on F32 kernel with:
```
Traceback (most recent call last):
  File "cgroup-limits", line 87, in <module>
    "NUMBER_OF_CORES": get_number_of_cores()
  File "cgroup-limits", line 74, in get_number_of_cores
    core_ids = list(map(int, group.split('-')))
ValueError: invalid literal for int() with base 10: '
```

hhorak: While I don't see it documented in https://www.kernel.org/doc/Documentation/cgroup-v2.txt, there is also `/sys/fs/cgroup/cpuset.cpus.effective`, and when running a container without CPU limit set, I see these values on cgroups v2:
```
# cat /sys/fs/cgroup/cpuset.cpus.effective
0-3
# cat /sys/fs/cgroup/cpuset.cpus

```

When running the container with cpus limited (`podman run -ti --rm --cpuset-cpus=1,2-3 fedora`), I see:
```
# cat /sys/fs/cgroup/cpuset.cpus.effective
1-3
# cat /sys/fs/cgroup/cpuset.cpus
1-3
```
So, `/sys/fs/cgroup/cpuset.cpus.effective` actually looks like the file we need to use for cgroups v2.

core/root/usr/bin/cgroup-limits
```diff
@@ -46,6 +46,8 @@ def get_memory_limit():
     """
 
     limit = _read_file('/sys/fs/cgroup/memory/memory.limit_in_bytes')
+    # If first file does not exist, try cgroups v2 file
+    limit = limit or _read_file('/sys/fs/cgroup/memory.max')
```
hhorak, line 50: When running a container with max memory set, both files `/sys/fs/cgroup/memory.max` on cgroups v2 returns the same value as `/sys/fs/cgroup/memory/memory.limit_in_bytes` on cgroups v1. However, when the memory is not limited for a container, `/sys/fs/cgroup/memory.max` returns 'max', while `/sys/fs/cgroup/memory/memory.limit_in_bytes` returns some crazy big number. I think the script should handle the `max` specifically and not print warning; instead, the max value should be used.

core/root/usr/bin/cgroup-limits
hhorak: Well, there is another issue, `/sys/fs/cgroup/cpuset.cpus.effective` nor `/sys/fs/cgroup/cpuset.cpus` don't exist when running a podman as non-root.

hhorak:
I've tried to address all the issues mentioned above inline, and added a test. A problem with the test and `cgroups-limits` generally is that not everything works when running podman as non-root. In that case, the test does not run all test cases.

frenzymadness:
Although I am not cgroups expert, your changes look good to me. I'd add one more test for limited cpus.

hhorak:
Good idea, the last commit adds a simple scenario test for `NUMBER_OF_CORES` value, but only uses 1 processor set, to make the test pass on even single-core VMs.

frenzymadness:
Fedora tests fail for an unrelated reason. The changes looks good to me.

PR #204, opened 2020-08-12
Add support for CentOS 8

hhorak:
While we have UBI8 images for s2i-core and s2i-base, we need to have centos 8 images as well for all the non-ubi images that will be built on top of those (mariadb, postgresql, etc).

### DIFF
--- a/base/Dockerfile.centos8
+++ b/base/Dockerfile.centos8
@@ -1,0 +1,50 @@
+# This image is the base image for all OpenShift v3 language container images.
+FROM centos/s2i-core-centos8
+
+ENV SUMMARY="Base image with essential libraries and tools used as a base for \
+builder images like perl, python, ruby, etc." \
+    DESCRIPTION="The s2i-base image, being built upon s2i-core, provides any \
+images layered on top of it with all the tools needed to use source-to-image \
+functionality. Additionally, s2i-base also contains various libraries needed for \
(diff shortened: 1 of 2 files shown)


hhorak:
[test]

frenzymadness:
[test]

hhorak:
> [test]

Tests for CentOS 8 are not yet ready in our CI infrastructure. It's WIP though. I'd probably leave this PR open till it's done (optimistic ETA: end of this week).

frenzymadness:
But it's listed here ^

hhorak:
> But it's listed here ^

I forgot that I used s2i-base as the testing material :) Good that it does something..

frenzymadness:
All green, I'd merge it!

phracek (approved):
I don't see any problematic parts. LGTM. Thanks for the PR!.

juliusvonkohout:
> While we have UBI8 images for s2i-core and s2i-base, we need to have centos 8 images as well for all the non-ubi images that will be built on top of those (mariadb, postgresql, etc).

Does this mean something like python-36-centos8 will also materialize soon?

pkubatrh:
>Does this mean something like python-36-centos8 will also materialize soon?

No, python-38 is a ubi8 image, hence no centos8 version is planned in the foreseeable future.

Issue #180, opened 2019-02-05
 get_number_of_cores function in cgroup-limits is wrong


atgreen:
The get_number_of_cores function currently reports the number of cores on the host, which is not what we want in a container.

The run-java.sh script in this s2i image does it right:  https://github.com/fabric8io-images/s2i/blob/master/java/images/rhel/run-java.sh#L147

yrro:
I guess this is the right link: https://github.com/fabric8io-images/run-java-sh/blob/7c1d123fb6e99182d08e20a21fb9513f248c45ed/fish-pepper/run-java-sh/fp-files/run-java.sh#L147

The logic (for cgroupsv1) is:

1. Read `/sys/fs/cgroup/cpu/cpu.cfs_quota_us`
2. If `-1`, there's no CPU limit; otherwise...
3. Read `/sys/fs/cgroup/cpu/cpu.cfs_period_us`
4. Number of cores is: `cfs_period_us`/`cfs_quota_us`

On my machine (with cgroupsv2):

1. Read two numbers (`max` and `period`) from from `/sys/fs/cgroup/cpu.max`
2. Number of cores is `max`/`period`

PR #209, opened 2021-05-05
Calculate available cores from cpu cgroups controller

yrro:
Fixes: https://github.com/sclorg/s2i-base-container/issues/180

### DIFF
--- a/core/root/usr/bin/cgroup-limits
+++ b/core/root/usr/bin/cgroup-limits
@@ -23,13 +23,14 @@ Variables currently supported:
         and from /sys/fs/cgroup/memory.max for cgroups v2
     NUMBER_OF_CORES
         Number of detected CPU cores that can be used. This value is
-        calculated from /sys/fs/cgroup/cpuset/cpuset.cpus for cgroups v1
-        and from /sys/fs/cgroup/cpuset.cpus.effective for cgroups v2
+        calculated from /sys/fs/cgroup/cpu/cpu.cfs_{quota,period}_us for
+        cgroups v1 and from /sys/fs/cgroup/cpuset.cpus.effective for cgroups v2
     NO_MEMORY_LIMIT
         Set to "true" if MEMORY_LIMIT_IN_BYTES is so high that the caller
         can act as if no memory limit was set. Undefined otherwise.
 """
 
+from __future__ import division
 from __future__ import print_function
 import sys
 
@@ -64,26 +65,35 @@ def get_number_of_cores():
     Read number of CPU cores.
     """
 
-    core_count = 0
+    quota, period = None, None
 
-    line = _read_file('/sys/fs/cgroup/cpuset/cpuset.cpus')
-    # If first file does not exist, try cgroups v2 file
-    line = line or _read_file('/sys/fs/cgroup/cpuset.cpus.effective')
-    if line is None:
-        # None of the files above exists when running podman as non-root,
-        # so in that case, this warning is printed every-time
-        print("Warning: Can't detect number of CPU cores from cgroups",
-              file=sys.stderr)
-        return None
+    quota = _read_file('/sys/fs/cgroup/cpu/cpu.cfs_quota_us')
+    if quota:
+        # cgroups v1
+        quota = quota.strip()
+        if quota == '-1':
+            quota = None
 
-    for group in line.split(','):
-        core_ids = list(map(int, group.split('-')))
-        if len(core_ids) == 2:
-            core_count += core_ids[1] - core_ids[0] + 1
-        else:
-            core_count += 1
+        period = _read_file('/sys/fs/cgroup/cpu/cpu.cfs_period_us')
+        if period:
+            period = period.strip()
 
-    return core_count
+    else:
+        # cgroups v2
+        line = _read_file('/sys/fs/cgroup/cpu.max')
+        if line:
+            fields = line.split()
+
+            if len(fields) >= 2:
+                quota = fields[0]
+                period = fields[1]
+
+    if quota.isdigit() and period.isdigit():
+        return int(quota)//int(period)
+
+    print("Warning: Can't detect number of CPU cores from cgroups",
+          file=sys.stderr)
+    return None
 
 
 if __name__ == "__main__":
--- a/core/root/usr/bin/cgroup-limits
+++ b/core/root/usr/bin/cgroup-limits
@@ -37,7 +37,6 @@ Variables currently supported:
 
 from __future__ import division
 from __future__ import print_function
-import multiprocessing
 import sys
 
 
@@ -71,7 +70,9 @@ def get_number_of_cores():
     Read number of CPU cores.
     """
 
-    limits = [_read_cpu_quota(), _read_cpuset_size(), multiprocessing.cpu_count()]
+    limits = [_read_cpu_quota(), _read_cpuset_size()]
+    if not limits:
+        return None
     return min([l for l in limits if l])
 
 
--- a/core/root/usr/bin/cgroup-limits
+++ b/core/root/usr/bin/cgroup-limits
@@ -22,9 +22,14 @@ Variables currently supported:
         /sys/fs/cgroup/memory/memory.limit_in_bytes for cgroups v1
         and from /sys/fs/cgroup/memory.max for cgroups v2
     NUMBER_OF_CORES
-        Number of detected CPU cores that can be used. This value is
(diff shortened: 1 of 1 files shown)


frenzymadness:
The code looks good to me and I've tested that it returns correct values for groups v1 and v2 and that it works for Python 3 and 2.

The only thing I'm curious about is why this problem has not been uncovered by CI? Let's try to run it - [test]

frenzymadness:
What's happening in the tests is a situation when quota is None and therefore has no method `isdigit`. You might use try/except and try to call int() on the variables and expect a ValueError for cases when it's not possible to transfer it to an integer and TypeError for cases when it's None. Or you can use something like `if quota and quota.isdigit() and …`.

yrro:
Well spotted--I think we may as well return `None` as soon as we determine there is no quota set -- there's no point going through the rest of the function. That also means we don't log the "unable to determine cpu quota" message when we _did_ determine that there is no quota.

[test]

yrro:
FYI, the old behaviour in the absence of limits was to query the `cpuset` cgroups controller. Should that behaviour be preserved?

frenzymadness:
[test]

frenzymadness:
I'm not an expert here. I'd like others to take a look as well. Cc @hhorak @pkubatrh @phracek @torsava

yrro:
The failure is...

```
  Testing 'NUMBER_OF_CORES' value...
NUMBER_OF_CORES=
test/run: line 34: [: : integer expression expected
NUMBER_OF_CORES not set.
S2I image 'rhscl/s2i-core-rhel7:1' test FAILED (exit code: 1)
```

... because in the absence of a cpu limit we are failing to output a `NUMBER_OF_CORES` value.

I'll wait for some feedback about whether the `cpuset` controller should be used as a fallback, or ignored, or somehow combined with the `cpu` controller to calculate this value...

phracek:
@gabemontero @lnykryn @kdudka I am not an expert with cgroups, do you have some in your team, who can help us? Any feedback is more then welcome :)

gabemontero:
> @gabemontero @lnykryn @kdudka I am not an expert with cgroups, do you have someone in your team, who can help us? Any feedback is more than welcome :)

Sorry @phracek no immediate recommendations or clear cut pointers to specific folks.  From an openshift team org perspective, perhaps the node team on CoreOS slack (#forum-node ... https://app.slack.com/client/T027F3GAJ/CK1AE4ZCK )

There is also openshift's current doc on node resources .... https://docs.openshift.com/container-platform/4.7/nodes/nodes/nodes-nodes-resources-configuring.html

Although not the same thing, about the closest thing I directly came to this was some JVM heap tuning for the Jenkins image that we borrowed from the the jboss/middleware folks.  But not close enough to draw any bits to this IMO.

@bparees @adambkaplan any additional thoughts?

bparees:
if there is no limit i'd be inclined to use the cpuset value since technically w/ no limit, all the cores are (potentially) available to you.

frenzymadness:
> if there is no limit i'd be inclined to use the cpuset value since technically w/ no limit, all the cores are (potentially) available to you.

That sounds reasonable to me…

phracek:
[test]

phracek:
Please use only double quotes and not mix with single quotes in python code.

phracek:
It looks like after modification by this PR the test suite does not work as expected.
See https://github.com/sclorg/s2i-base-container/blob/master/core/test/run#L34

Log from test:
```bash
  Testing 'NUMBER_OF_CORES' value...
NUMBER_OF_CORES=
test/run: line 34: [: : integer expression expected
NUMBER_OF_CORES not set.
```

In some cases, the python script returns None and the bash test fails. Can you please fix it as well?

phracek:
[test]

yrro:
> if there is no limit i'd be inclined to use the cpuset value since technically w/ no limit, all the cores are (potentially) available to you.

How about getting the number of CPUs from both the cpu (quota) and cpuset controllers, and using whichever value is lower?

This way if a container runtime only configures either controller, the value is correct.

I'm trying to think of scenarios where you'd run a container with both controllers configured, and I've come up with:

* I have a dedicated set of CPUs for, say, running databases (0-4) AND I want to run individual database containers with a limit of 2 cores each - I would expect `NUMBER_OF_CORES=2`

I can't think of a useful case for running a container with a cpuset of size 2 and a cpu quota of 4, so using the lower of the two values in this case seems harmless.

I also added a fallback to return the number of CPUs present in the machine. But maybe that's beyond the scope of a script called `cgroup-limits`; I'm happy to remove that if you think it's not desirable.

yrro:
Tests are failing with:

```
11:14:15 -> Pulling image centos:8 before building image from Dockerfile.centos8.
11:14:15 Trying to pull repository docker.io/library/centos ... 
11:14:15 toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

Maybe you could use `quay.io/centos/centos` instead?

phracek:
Thank you for this catch.

There are two options for how to fix it. Either as a separate PR or by this PR.
Fixing by this pull request does make sense for me. Can you please do it?

yrro:
Sure - see https://github.com/sclorg/s2i-base-container/pull/209

yrro:
Sorry I just realised you're happy fixing it here - I'll do that then.

yrro:
Sorry I just realised you were ok to do it here. I'll leave that other PR open anyway & rebase this one on top once it's merged.

phracek:
[test]

phracek:
Let's try one more time. Infra problems.

[test]

yrro:
`centos8` still failing with

```
12:18:04 Trying to pull repository docker.io/library/centos ... 
12:18:54 toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
12:18:54 Pulling image centos:8 failed.
```

I think I need to rebase this PR--doing that now...

phracek:
[test]

yrro:
I've thought about the use of {{multiprocessing}} a bit more and I think it's better to remove the fallback to number of cores.

If I'm deploying a webapp to my OpenShift cluster and forget to set a CPU limit, I'd rather have it fall back to running a single request handler, rather than have it spawn 40 of them and end up using lots of memory unintentionally.

yrro:
That said, `cpuset.cpus.effective` will still contain `[0-39]` in this case. So the fall back to using `multiprocessing` is never used anyway.

yrro:
[test]

yrro:
Guess I can't trigger the tests myself :)

phracek:
> Guess I can't trigger the tests myself :)

No, only owners of the repository

[test]

yrro:
Argh this time it's

```
11:54:58 Step 1/5 : FROM centos/s2i-core-centos8
11:54:58 Trying to pull repository docker.io/centos/s2i-core-centos8 ... 
11:54:58 toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

While building `base/Dockerfile.centos8`.

`base/Dockerfile` succeeds because it's built `FROM quay.io/centos7/s2i-core-centos7` but there doesn't appear to be an equivalent for the centos 8 CI to use.

yrro:
Who can we ask to get a mirror of `s2i-core-centos8` on quay.io?

phracek:
> Who can we ask to get a mirror of `s2i-core-centos8` on quay.io?

I guess a responsible person who owns quay.io/centos organization.

phracek:
[test-all]

yrro:
As of 4f4f13847fd372c304503bc0bf8e11fae6ce9e06, `core/root` is now a symlink into `common` which is a submodule.

I've opened a new PR against the submodule's repo: https://github.com/sclorg/container-common-scripts/pull/237

yrro:
Any chance you can take a look at https://github.com/sclorg/container-common-scripts/pull/237 @phracek?

zmiklank:
Hi, if I understand this correctly, these changes are already present in the container-common-scripts repository. Can this PR be closed?

hhorak:
> Hi, if I understand this correctly, these changes are already present in the container-common-scripts repository. Can this PR be closed?

Let's close it. I'll open a new one to address cgroup-limits script to fail with python traceback when running a container as non-root, but it's not directly related to the changes here.

PR #213, opened 2021-09-23
Pull from registry.centos.org instead of (implicit) Docker Hub

yrro:
Bypasses the pull rate-limit.

### DIFF
--- a/core/Dockerfile.centos8
+++ b/core/Dockerfile.centos8
@@ -1,5 +1,5 @@
 # This image is the base image for all s2i configurable container images.
-FROM centos:8
+FROM registry.centos.org/centos/centos:8
 
 ENV SUMMARY="Base image which allows using of source-to-image."	\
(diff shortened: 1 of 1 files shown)


core/Dockerfile.centos8
```diff
@@ -1,5 +1,5 @@
 # This image is the base image for all s2i configurable container images.
-FROM centos:8
+FROM registry.centos.org/centos/centos:8
```
yrro, line 2: The official https://centos.org/centos-linux/ page still points people to the Docker Hub. But this registry is already used in `core/Dockerfile.centos7`.

`base/Dockerfile.centos8` also has an unqualified reference to `centos/s2i-core-centos8` but I don't see a mirror of it in quay.io (whereas `base/Dockerfile.centos7` uses `quay.io/centos7/s2i-core-centos7`).

phracek:
[test]

phracek:
The image itself is pull able. I don't see any reason, why not to merge it.

PR #215, opened 2021-09-24
Move shell scripts from core to common-scripts

frenzymadness:
This is important for minimal images not based on s2i-core.

Related to: https://github.com/sclorg/s2i-python-container/pull/467 has to be merged together with: https://github.com/sclorg/container-common-scripts/pull/210


frenzymadness:
@hhorak @pkubatrh @phracek @torsava  I need to know your opinion here because I'm probably far away from all the use cases and all the tools working with those container images. The final image should be the same, no problem there. Building with `make build` also works because the docker context folder is one level higher.

We are using something very similar for Python containers for a long time and it works. Every time I copy something from the upstream repository to dist-git, I'm using -L for the cp command so it follows symlinks and does not copy them as they are.

If you want to play with it, switch your common scripts to https://github.com/sclorg/container-common-scripts/tree/core-scripts

hhorak:
The only thought I have is about the name `container-root` -- that directory is gonna be added as a whole usually. This change opens an opportunity to share some other scripts this way for other purposes, e.g. between mariadb and mysql or for containers based on s2i-base now, but building on a different parent.. (just random use cases, that are likely not real now).. so the point is that we can call it rather `core-root`, so we can add more such "roots" in the future that will be called differently.

pkubatrh:
Maybe even add another directory layer (something like `shared-scripts/core` instead of just `core-root`) if we are already planning to share some others.

From the tools viewpoint, I think having the scripts in common-scripts should be ok - we already have other scripts in there so the tooling should be prepared to copy them in if needed.

frenzymadness:
Thanks for the feedback. I've renamed the folder in both PRs.

frenzymadness:
This is no longer a draft. All feedbacks are incorporated and I believe it's ready to be merged.

frenzymadness:
[test]

frenzymadness:
I was afraid that I forget to change relative paths in the base image but there are no copy commands so the changed docker context should have zero impact there.

frenzymadness:
The tests are running for 9 days now. There is a conflict in the common module which is okay, this should be merged after we merge https://github.com/sclorg/container-common-scripts/pull/210 so this PR can contain also the latest common submodule with the new folder.

[test]

frenzymadness:
Since the common scripts submodule is ready for this change, I've updated it and did a rebase. This should be now ready for a merge.

[test]

frenzymadness:
All green!

PR #219, opened 2021-12-15
Add support for CentOS Stream 9

phracek:
This commit adds support for building s2i-core and s2i-base containers in CentOS Stream 9.

This Pull request depends on https://github.com/sclorg/container-common-scripts/pull/224

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -6,9 +6,25 @@ on:
 
 jobs:
   build-and-push:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - dockerfile: "Dockerfile"
+            registry_namespace: "centos7"
+            os_type: "centos7"
+            registry_suffix: "-centos7"
+          - dockerfile: "Dockerfile.c9s"
+            registry_namespace: "sclorg"
+            os_type: "c9s"
+            registry_suffix: ""
+
     steps:
       - uses: actions/checkout@v2
+        -
+      - name: Set current date as env variable
+        run: echo "NOW=$(date +'%Y%m%d')" >> $GITHUB_ENV
 
       - name: Get base image name
         id: base-image-name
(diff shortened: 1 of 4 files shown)


hhorak:
Same comment as in https://github.com/sclorg/container-common-scripts/pull/224#issuecomment-1005744792 -- I'd prefer `c9s` instead of `stream9` because it is more common already.

base/Dockerfile.stream9
```diff
@@ -0,0 +1,49 @@
+# This image is the base image for all OpenShift v3 language container images.
+FROM quay.io/sclorg/s2i-core-stream9
```
hhorak, line 2: For s2i-python-minimal, we decided to use `:el8` as a tag instead of using the el version in the image name. Couldn't we use `quay.io/sclorg/s2i-core:c9s` here then?

core/Dockerfile.stream9
```diff
@@ -0,0 +1,73 @@
+# This image is the base image for all s2i configurable container images.
+FROM quay.io/centos/centos:stream9
```
hhorak, line 2: Now I see where `stream9` is coming from. Being it used close to "centos" name, it makes it quite clear, I'm still worried about being it used without centos context as mentioned elsewhere (that `c9s` might be more clear).

phracek: Let's use c9s tag. I totally agree.

base/Dockerfile.stream9
phracek: Make sense to me.

.github/workflows/build-and-push-stream9.yml
```diff
@@ -0,0 +1,67 @@
+name: Push CentOS Stream 9 images to Quay.io registry
```
zmiklank, line 1: The `build-and-push-stream9.yml` and `build-and-push.yml` are quite similar GitHub actions. Wouldn't it be possible to use a matrix here, so no duplicated code would originate?
Matrix could be (IMHO) consisted of "registry", "dockerfiles" and "image" entries.

phracek: This is a really good point. Let's use matrix.

phracek: Fixed by https://github.com/sclorg/s2i-base-container/pull/219/commits/0745d917e0995b94645d242aa14d50a80d1b53a6

core/Dockerfile.stream9
phracek: Fixed by https://github.com/sclorg/s2i-base-container/pull/219/commits/0745d917e0995b94645d242aa14d50a80d1b53a6

base/Dockerfile.stream9
phracek: Fixed by https://github.com/sclorg/s2i-base-container/pull/219/commits/0745d917e0995b94645d242aa14d50a80d1b53a6

phracek:
[test]

phracek:
[test]

.github/workflows/build-and-push.yml
```diff
@@ -59,9 +87,9 @@ jobs:
         with:
           image: ${{ steps.build-image-base.outputs.image }}
           tags: ${{ steps.build-image-base.outputs.tags }}
-          registry: quay.io/centos7
-          username: ${{ secrets.QUAY_IMAGE_BUILDER_USERNAME }}
-          password: ${{ secrets.QUAY_IMAGE_BUILDER_TOKEN }}
+          registry: quay.io/{{ matrix.registry_namespace }}
+          username: ${{ env.USERNAME }}
```
hhorak, line 91: I'm not familiar with the actions enough to be able to tell whether exposing password/token to the environment has some security implications -- like if it would be easier to gain the token via a nifty change in a pull-request... I'm thinking whether the password couldn't be only used only as a matrix variable:
```
      matrix:
        include:
          - dockerfile: "Dockerfile"
            registry_namespace: "centos7"
            os_type: "centos7"
            registry_suffix: "-centos7"
            password: "${{ secrets.QUAY_IMAGE_BUILDER_USERNAME }}"
```
and then used as `${{ matrix.password }}`

That's the only comment I have, otherwise I haven't spotted anything suspicious, but must again repeat that I'm not any expert in the github actions.

phracek: This is not possible see article here: https://github.community/t/secrets-in-matrix/17307

PR #223, opened 2022-01-07
Add missing root dir

phracek:


phracek:
The same as here: https://github.com/sclorg/s2i-base-container/blob/master/core/Dockerfile.fedora#L55

PR #227, opened 2022-01-11
Do not use symlink, but rather path from common

phracek:
During the building by GitHub Actions, the file /common/shared-scripts/core is not found.

This should fix the problem.

See error:

```bash
6 [internal] load build context
#6 transferring context: 87B done
#6 DONE 0.0s

#7 [3/5] COPY ./core/root/ /
#7 ERROR: "/common/shared-scripts/core": not found

#4 [1/5] FROM quay.io/centos/centos:stream9@sha256:28356bcd2b62bf775a344603d22c1d29c4a855ff4d4e853558ee72b20952ffb8
#4 resolve quay.io/centos/centos:stream9@sha256:28356bcd2b62bf775a344603d22c1d29c4a855ff4d4e853558ee72b20952ffb8 done
#4 sha256:6e0e003895ae89f80beb974738572d53516e97ff1503f572d37a4402cb2e7913 0B / 55.26MB
#4 DONE 0.0s

#5 [2/5] RUN INSTALL_PKGS="bsdtar   findutils   groff-base   glibc-locale-source   glibc-langpack-en   gettext   rsync   scl-utils   tar   unzip   xz   yum" &&   mkdir -p /opt/app-root/src/.pki/nssdb &&   chown -R 1001:0 /opt/app-root/src/.pki &&   yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS &&   rpm -V $INSTALL_PKGS &&   yum -y clean all --enablerepo='*'
#5 CANCELED
```

The link is here: https://github.com/sclorg/s2i-base-container/runs/4776067753?check_suite_focus=true

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -58,7 +58,7 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
   yum -y clean all --enablerepo='*'
 
 # Copy extra files to the image.
-COPY ./core/root/ /
+COPY ./common/shared-scripts/core/ /
 
 # Directory with the sources is set as the working directory so all STI scripts
 # can execute relative to this path.
--- a/core/Dockerfile.c9s
+++ b/core/Dockerfile.c9s
@@ -57,7 +57,7 @@ RUN INSTALL_PKGS="bsdtar \
   yum -y clean all --enablerepo='*'
 
 # Copy extra files to the image.
-COPY ./core/root/ /
+COPY ./common/shared-scripts/core/ /
 
 # Directory with the sources is set as the working directory so all STI scripts
 # can execute relative to this path.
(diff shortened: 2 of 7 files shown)


phracek:
@frenzymadness Can you please take a look at it?

phracek:
[test]

frenzymadness:
I'm afraid that this won't work in downstream dist-gits where we don't have the common submodule. The symlink is there so if you copy the core with `-L` option, `cp` follows symlinks and copies the important files to the right location in the core folder so basically, downstream does not have the common submodule as a whole but has some important files from it thanks to the symlinks like this one.

phracek:
@frenzymadness Unfortunately command COPY (https://docs.docker.com/engine/reference/builder/#copy) does not have option -L. I don't follow your point.

phracek:
Submodules were not checkout.

PR #229, opened 2022-01-12
Let's move back to redhat-actions

phracek:
We are moving back to redhat cations. docker/login-action does not work properly as we want.

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -15,9 +15,11 @@ jobs:
           - dockerfile: "Dockerfile"
             registry_namespace: "centos7"
             registry_suffix: "-centos7"
+            tag: "centos7"
           - dockerfile: "Dockerfile.c9s"
             registry_namespace: "sclorg"
             registry_suffix: ""
+            tag: "c9s"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-push.yml
```diff
@@ -37,49 +35,71 @@ jobs:
...
-        id: buildx
-        uses: docker/setup-buildx-action@v1
+      - name: Build s2i-core image
+        id: build-image-core
+        # https://github.com/marketplace/actions/buildah-build
+        uses: redhat-actions/buildah-build@v2
+        with:
+          dockerfiles: core/${{ matrix.dockerfile }}
+          image: ${{ steps.base-image-name.outputs.core_name }}
+          tags: latest c9s ${{ env.NOW }}-${{ github.sha }}
```
zmiklank, line 49: Does this mean, that also centos7 images will be tagged with c9s?

phracek: Good catch

zmiklank:
Thanks for PR, it LGTM now.

zmiklank (approved):

PR #238, opened 2022-03-14
Let's build s2i-core and s2i-base container each Tuesday at 1:00

phracek:
This pull request builds s2i-core and s2i-base container each Tuesday at 1:00

@hhorak @frenzymadness @zmiklank Can you please take a look at it and review it?

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -3,7 +3,10 @@ on:
   push:
     branches:
       - master
-
+    scheduled:
+      # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule
+      # Schedule build to 1:00 each Tuesday
+      - cron: '0 1 * * 2'
 jobs:
   build-and-push:
     name: Build and push s2i containers


zmiklank:
Sorry for being lost, but why is this needed, and why rebuild specifically on Tuesdays? Is it a part of a chain-rebuilds?

Otherwise, it seems to me that the PR is functional.

zmiklank:
Sorry for being lost, but why is this needed, and why rebuild specifically on Tuesdays? Is it a part of a chain-rebuilds?

Otherwise, it seems to me that the PR is functional.

phracek:
> Sorry for being lost, but why is this needed, and why rebuild specifically on Tuesdays? Is it a part of a chain-rebuilds?
> 
> Otherwise, it seems to me that the PR is functional.

During Friday's meeting with @frenzymadness @hhorak @pkubatrh We have decided, that we can use GitHub Action to do "chain rebuilds". It means, during Tuesday, we will build s2i-core and s2i-base containers, because they are in the same repository, and the day after or a couple of hours after s2i-core and s2i-base rebuilds we will initiate rebuilds of all our containers across Fedora, CentOS7, and CentOS Stream 9.

zmiklank:
Oh, thanks for the explanation! :)

frenzymadness:
Looks good to me, let's verify that it works and then we can add dependant container images and see that they will build on top of this every week. Good job. Thank you!

PR #252, opened 2022-05-14
Remove pointless comment about prepare-yum-repositories

frenzymadness:
The script is actually copied and used only on RHEL 7.


phracek:
[test]

phracek (changes requested):
Thanks for the PR. Actually we have Dockerfile.rhel9 files https://github.com/sclorg/s2i-base-container/blob/master/core/Dockerfile.rhel9#L31. Please remove this change as well.

frenzymadness:
> Thanks for the PR. Actually we have Dockerfile.rhel9 files https://github.com/sclorg/s2i-base-container/blob/master/core/Dockerfile.rhel9#L31. Please remove this change as well.

Removed. Please merge it before we add RHEL 10 :smile:

phracek:
[test]

PR #254, opened 2022-05-16
Fixes #253: Support for Fedora 36

### LINKED ISSUE #253: Fedora 36
Fedora 36 is out so we can update. There is no rush as Fedora 35 EOL will be in November. From Python perspective, Fedora 36 still has Python 3.10 as the main Python interpreter so there is currently no reason for updating. What about other images?

phracek:
This pull request adds support for Fedora 36 into s2i-core and s2i-base images

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -35,7 +35,7 @@ jobs:
             suffix: "c8s"
           - dockerfile: "Dockerfile.fedora"
             registry_namespace: "fedora"
-            tag: "35"
+            tag: "36"
             quayio_username: "QUAY_IMAGE_FEDORA_BUILDER_USERNAME"
             quayio_token: "QUAY_IMAGE_FEDORA_BUILDER_TOKEN"
 
--- a/.github/workflows/container-tests.yml
+++ b/.github/workflows/container-tests.yml
@@ -13,10 +13,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - tmt_plan: "c8s"
+          - tmt_plan: "fedora"
             os_test: "fedora"
             context: "Fedora"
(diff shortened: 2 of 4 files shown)


frenzymadness (commented):
The changes look good to me. I have just one concern. If we switch now, we stop building Fedora 35, right? I mean, when we were just an upstream and Fedora images have been built on Fedora infra, it didn't matter because we can maintain parallel versions in downstream. But now, Fedora images are produced here and pushed to quay.io which means that Fedora 35 will soon be outdated.
This does not really matters for packages installed on higher levels (like Python or Perl) because they are updated during every rebuild but what about components in base and core container images?
Maintaining two or more Fedora versions does not make sense to me, it's too much work for little benefit so we can maybe wait for a while with the update. What do you think?

phracek:
Yes, of course, switching to F36, we will stop building it to quay.io/fedora registries. But I guess, if a rebuild needs to be done, we can do it manually.
But quay.io/fedora/s2i-core:35 remains in the registry.

The change itself would be needed if packages or variables will be added to the s2i-core, respectively s2i-base image.
Then we have to rebuild the rest of the images as well.

For now maintaining s2i-core, respectively s2i-base images are hard and do not bring so much benefit.

Personally, I don't see any issues, for now, we add more packages, variables etc.

Of we can update s2i-core and s2i-base images, like bootstrap:
1) add variables, and packages into Dockerfiles for supported versions, like F36
2) Switch quay.io/fedora/fedora:35 to quay.io/fedora/fedora:34 and build them
3) Switchback or revert commit.

Or add more docker files, or even, use dist-gen and @phracek is hiding for this sentence :) and forget about it.

phracek:
@frenzymadness Can we merge it?

frenzymadness (approved):

PR #258, opened 2022-07-19
build_and_push: detect fedora tag from Dockerfile

zmiklank:
Detect automatically fedora tag from Dockerfile in build_and_push_action.
The goal is to automatically detect change between Fedora version, e.g.,
35->36.

The only place, from where version transition can be spotted is
Dockerfile.fedora.

The used tag has to be in format "36".

curl for getting the Dockerfile.fedora was used because we do not have
repository cloned at that point in the action.
Cloning the whole repository because of one Dockerfile would be ineffective.

Should fix https://github.com/sclorg/build-and-push-action/issues/14.

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -37,11 +37,23 @@ jobs:
             suffix: "c8s"
           - dockerfile: "Dockerfile.fedora"
             registry_namespace: "fedora"
-            tag: "36"
             quayio_username: "QUAY_IMAGE_FEDORA_BUILDER_USERNAME"
             quayio_token: "QUAY_IMAGE_FEDORA_BUILDER_TOKEN"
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-push.yml
```diff
@@ -37,11 +37,24 @@ jobs:
             suffix: "c8s"
           - dockerfile: "Dockerfile.fedora"
             registry_namespace: "fedora"
-            tag: "36"
+            tag: "fedora"
```
pkubatrh, line 40: Do we have to keep this even when it is not used for the Fedora images?

zmiklank: No. I have removed it.

pkubatrh:
Otherwise LGTM.

PR #261, opened 2022-11-15
Update to F37

hhorak:


hhorak:
[test]

phracek:
RHEL9 test failed on:
```bash
STEP 5/11: RUN INSTALL_PKGS="bsdtar   findutils   groff-base   glibc-locale-source   glibc-langpack-en   gettext   rsync   scl-utils   tar   unzip   xz   yum" &&   mkdir -p ${HOME}/.pki/nssdb &&   chown -R 1001:0 ${HOME}/.pki &&   yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS &&   rpm -V $INSTALL_PKGS &&   yum -y clean all --enablerepo='*'
/bin/sh: error while loading shared libraries: libc.so.6: cannot change memory protections
Error: error building at STEP "RUN INSTALL_PKGS="bsdtar   findutils   groff-base   glibc-locale-source   glibc-langpack-en   gettext   rsync   scl-utils   tar   unzip   xz   yum" &&   mkdir -p ${HOME}/.pki/nssdb &&   chown -R 1001:0 ${HOME}/.pki &&   yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS &&   rpm -V $INSTALL_PKGS &&   yum -y clean all --enablerepo='*'": error while running runtime: exit status 127
make[1]: Leaving directory '/tmp/sclorg/s2i-base-container'
```

This should be filed. @hhorak Can you please do it?

base/Dockerfile.fedora
```diff
@@ -1,5 +1,5 @@
 # This image is the base image for all OpenShift v3 language container images.
-FROM quay.io/fedora/s2i-core:36
+FROM quay.io/fedora/s2i-core:37
```
zmiklank, line 2: Is there any reason for VERSION in base image being 1?

hhorak: Good question, I guess your suggestion is changing it to the same as core has, right? It seems like a good idea to me.

zmiklank: If there is not any reason for keeping the 1 there, then yes, I would suggest to change it to 37, same as for s2i-core.

hhorak: Changing the version does not have impact on the resulting image tag though: https://github.com/sclorg/s2i-base-container/blob/master/.github/workflows/build-and-push.yml#L50
...but we can do it still.

hhorak: Instead of re-defining it, I've removed VERSION=... entirely from the s2i-base, so it inherits VERSION from the s2i-core parent image.

hhorak:
[test]

hhorak:
The only comment addressed, CI green, I guess we're good to go.

PR #262, opened 2022-12-21
Update common submodule

zmiklank:
And update also gitignore


zmiklank:
Fixes: https://github.com/sclorg/postgresql-container/issues/482

PR #265, opened 2023-05-02
bump build-and-push action version, remove deprecated set-output

zmiklank:


phracek (commented):
I would add an explanation, of why `-` was added. At least in the commit message. And maybe an explanation, that GA build-and-push-action changes an API.

zmiklank:
Good point, thanks. The explanation was added.

phracek (approved):
Thanks for the comment. Awesome

PR #266, opened 2023-05-19
Multi arch support for s2i-core-c9s

modassarrana89:
Updated build and push  yaml file 
1. conditional check for s2i-core image built only for dockerfile.c9s
2. Pass arch option for multi-arch image built.

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -53,14 +53,28 @@ jobs:
             _tag=${{ matrix.tag }}
           fi
           echo "tag=${_tag}" >> $GITHUB_OUTPUT
-
-      - name: Build and push s2i-core to quay.io registry
+          
+      - name: Check if dockerfile is c9s , then build & push s2i-core for multi-arch
+        if: matrix.dockerfile == 'Dockerfile.c9s'
         uses: sclorg/build-and-push-action@v4
         with:
           registry: "quay.io"
           registry_namespace: ${{ matrix.registry_namespace }}
-          registry_username: ${{ secrets[matrix.quayio_username] }}
-          registry_token: ${{ secrets[matrix.quayio_token] }}
+          registry_username: ${{ matrix.quayio_username }}
+          registry_token: ${{ matrix.quayio_token }}
+          dockerfile: "core/${{ matrix.dockerfile }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-push.yml
```diff
@@ -53,8 +53,22 @@ jobs:
...
+        uses: sclorg/build-and-push-action@v4
+        with:
+          registry: "quay.io"
+          registry_namespace: ${{ matrix.registry_namespace }}
+          registry_username: ${{ secrets[matrix.quayio_username] }}
+          registry_token: ${{ secrets[matrix.quayio_token] }}
+          dockerfile: "core/${{ matrix.dockerfile }}"
+          image_name: "s2i-core${{ matrix.suffix }}"
+          tag: ${{ steps.tag.outputs.tag }}
+          archs: amd64, ppc64le, s390x
```
zmiklank, line 68: What if we specify only `ppc64le, s390x` here and build the `amd64` with during next step with other images? Could that work?
```suggestion
          archs: ppc64le, s390x
```

.github/workflows/build-and-push.yml
```diff
@@ -53,8 +53,22 @@ jobs:
...
 
-      - name: Build and push s2i-core to quay.io registry
+      - name: if dockerfile is not c9s , Build and push s2i-core to quay.io registry
+        if: matrix.dockerfile != 'Dockerfile.c9s'
```
zmiklank, line 71:
```suggestion

```

.github/workflows/build-and-push.yml
modassarrana89: let me check , will update you post testing

modassarrana89: As conveyed , it didn't work that way . I have to add that old check if: matrix.dockerfile != 'Dockerfile.c9s' as well , otherwise it will override the tag again for c9s 
https://github.com/modassarrana89/s2i-base-container/actions/runs/5042239798/jobs/9042671876

.github/workflows/build-and-push.yml
modassarrana89: This change is needed else image tag for c9s will get override with amd64 only

zmiklank: Ok, understood. Thank you for testing this approach.

PR #278, opened 2023-08-30
Add support for building s2i-core and s2i-base from Fedora 39

phracek:

### DIFF
--- a/base/Dockerfile.fedora
+++ b/base/Dockerfile.fedora
@@ -1,5 +1,5 @@
 # This image is the base image for all OpenShift v3 language container images.
-FROM quay.io/fedora/s2i-core:38
+FROM quay.io/fedora/s2i-core:39
 
 ENV SUMMARY="Base image with essential libraries and tools used as a base for \
 builder images like perl, python, ruby, etc." \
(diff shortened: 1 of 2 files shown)


phracek:
[test]

zmiklank:
Is there any reason for moving core and base to f39 just now?

If not, I would rather leave the default one being 38, as main part of our Fedora images is based on 38 and with this change the 38 image will stop getting new updates with rebuilds disabled.

phracek:
@zmiklank @jplesnik is planning to add new Perl-5.38 version that exists only since Fedora 39. Pull request will be available soon.
Only from this reason, I have created this pull request.

phracek:
And python-3.12 also needs Fedora 39.

frenzymadness:
According to the discussion we had at the meeting this week, we should think about how to make it possible to have more than one Fedora version here.

phracek:
q

> According to the discussion we had at the meeting this week, we should think about how to make it possible to have more than one Fedora version here.

I guess, this can be done as separate pull request, right? The issue is here: https://github.com/sclorg/s2i-base-container/issues/279

zmiklank:
Closing in favor of https://github.com/sclorg/s2i-base-container/pull/280.

PR #282, opened 2023-09-27
Do not use default tags for built f39 base and core image

zmiklank:


frenzymadness:
I like it. One side note from me is that we might want to release a new version of the action and require that version here so we are sure that the functionality is available. But it's not a hard requirement.

Issue #283, opened 2023-11-08
Investigate /sbin/nologin as default shell for default user, and change it if needed


zmiklank:
We came to conclusion that the /sbin/nologin as a default shell for a default container user does not have a real reason now.
Even though it does not really break things (with one exception that we know of) we would want to remove this setting as it is confusing to have it there without a reason.

PR #284, opened 2023-11-09
drop /sbin/nologin as default shell for default user

zmiklank:
Only known use-cases not working when /sbin/nologin is set as default shell are:
1) executing $SHELL explicitly in already running container
2) explicitly running 'login' command when starting the container (e.g.,
   podman run --rm <imagename> login)
3) connecting to running container with 'exec login'
4) ssh-ing to container

And there is not any reason why these use-cases should not work in sclorg images.

The openshift runs the containers under random UID. In this case the default shell is /bin/sh. This removal thus should not change the behavior of the images run on openshift.

Further changes will be needed in images using NSS_WRAPPER and in micro images because the user is generated differently there.

Partly closes: https://github.com/sclorg/s2i-base-container/issues/283

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -69,6 +69,5 @@ CMD ["base-usage"]
 
 # Reset permissions of modified directories and add default user
 RUN rpm-file-permissions && \
-  useradd -u 1001 -r -g 0 -d ${HOME} -s /sbin/nologin \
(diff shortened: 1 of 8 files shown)


zmiklank:
[test]

phracek (approved):
LGTM. I did not hit any issue. Test are completely passing.

PR #285, opened 2023-11-16
Switch using NodeJS from version 14 to version 20

phracek:
Switch using NodeJS from version 14 to version 20

NodeJS version 14 reached EOL and version 20
was already introduced.

Also Dockerfile's checks if node -v corresponds with RHEL8 and RHEL9 version

```bash
$ node -v 
v20
$
```

### DIFF
--- a/base/Dockerfile.c8s
+++ b/base/Dockerfile.c8s
@@ -7,7 +7,7 @@ builder images like perl, python, ruby, etc." \
 images layered on top of it with all the tools needed to use source-to-image \
 functionality. Additionally, s2i-base also contains various libraries needed for \
 it to serve as a base for other builder images, like s2i-python or s2i-ruby." \
-    NODEJS_VER=14
+    NODEJS_VER=20
 
 LABEL summary="$SUMMARY" \
       description="$DESCRIPTION" \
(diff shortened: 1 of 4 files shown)


phracek:
[test]

zmiklank (changes requested):
Hi.
This solution does not work for RHEL9, as it does not use nodejs from module (yet).
Thanks for you work!.

phracek:
[test]

hhorak:
> This solution does not work for RHEL9, as it does not use nodejs from module (yet).

nodejs:20 is a module in RHEL-9 actually, so it should work the same as in RHEL-8.

zmiklank:
> > This solution does not work for RHEL9, as it does not use nodejs from module (yet).
> 
> nodejs:20 is a module in RHEL-9 actually, so it should work the same as in RHEL-8.

@hhorak Right, but when I commented on this PR the module was not enabled in the Dockerfile yet. Peto already fixed it.

zmiklank:
LGTM for RHELs. What about other distros? Do we want to keep there the nodejs version which has been there already?

zmiklank:
[test]

zmiklank (approved):
LGTM. Thanks. Feel free to merge when tests are green :).

PR #288, opened 2024-02-12
Move latest fedora release to f40

zmiklank:

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -33,9 +33,9 @@ jobs:
             suffix: "-c8s"
             use_default_tags: 'true'
 
-          - dockerfile: "Dockerfile.f38"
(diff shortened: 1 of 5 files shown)


hhorak:
LGTM, but I might not realize all the consequences on TF side, so take my feedback with some respect to not be complete ;)

frenzymadness:
Isn't it too soon? I mean, branching for F40 is today and the target date for release is in the middle of April.

zmiklank:
We need f40 for ruby 3.3 and mariadb 10.11.
I think it would be beneficial for us to start to use and test these images rather sooner, also because of future RHEL images.
But this is planned to be discussed further tomorrow on the container meeting.

zmiklank:
Closing in favor of: https://github.com/sclorg/s2i-base-container/pull/289.

PR #289, opened 2024-02-14
add support for f40

zmiklank:
Start building and pushing to quay.io s2i-base and s2i-core images for f40.
We planned to support 2 Fedora version simultaneously, but this decision has been revised.
Only tested version is still f38.

### DIFF
--- a/core/Dockerfile.f40
+++ b/core/Dockerfile.f40
@@ -1,0 +1,70 @@
+# This image is the base image for all s2i configurable container images.
+FROM quay.io/fedora/fedora:40
+
+ENV SUMMARY="Base image which allows using of source-to-image."	\
+    DESCRIPTION="The s2i-core image provides any images layered on top of it \
+with all the tools needed to use source-to-image functionality while keeping \
+the image size as small as possible." \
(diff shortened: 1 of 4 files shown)


core/Dockerfile.f40
```diff
@@ -0,0 +1,70 @@
+# This image is the base image for all s2i configurable container images.
+FROM quay.io/fedora/fedora:40
+
+ENV SUMMARY="Base image which allows using of source-to-image."	\
+    DESCRIPTION="The s2i-core image provides any images layered on top of it \
+with all the tools needed to use source-to-image functionality while keeping \
+the image size as small as possible." \
+    NAME=s2i-core \
+    VERSION=38 \
```
phracek, line 9: Typo here.

phracek (changes requested):
Please address my comments

core/Dockerfile.f40
zmiklank: Good catch. The same problem is with f39. Will fix that.

phracek (approved):

PR #291, opened 2024-02-16
s2i-base, f40: install available package alternatives

zmiklank:
@ljavorsk Can you please confirm, that the proper way how to handle zlib to zlib-ng migration is the one used in this PR?


ljavorsk:
The correct one is `zlib-ng-compat-devel`, beside that LGTM

PR #292, opened 2024-03-06
build s2i-[base,core] for arm too

zmiklank:


frenzymadness (approved):
So, we are doing more with less code, great. I really like it!

Issue #294, opened 2024-03-19
CentOS Stream official images on quay.io not accesible for s390x


zmiklank:
Because of that our build-and-push action is failing.
We should remove this one architecture from a list of build targets for s2i-base and s2i-core.

PR #295, opened 2024-03-19
c8s does not provide s390x arch, stop these builds

zmiklank:
Fixes: https://github.com/sclorg/s2i-base-container/issues/294

Fixes:
```
STEP 1/10: FROM quay.io/centos/centos:stream8
Trying to pull quay.io/centos/centos:stream8...
error creating build container: choosing an image from manifest list docker://quay.io/centos/centos:stream8: no image found in image index for architecture s390x, variant "", OS linux
```


phracek (approved):
Thanks LGTM.

PR #308, opened 2024-07-15
ci: enable Automatic Merge of Approved Pull Requests

phracek:
Add automation that will merge approved PRs with passed CI checks automatically.

It uses the following GitHub Actions:

* [gather-pull-request-metadata](https://github.com/redhat-plumbers-in-action/gather-pull-request-metadata) - to gather important data about PRs (used in following actions)
* [pull-request-validator](https://github.com/redhat-plumbers-in-action/pull-request-validator) - to check if PR is approved and has passed CI checks and more
* [auto-merge](https://github.com/redhat-plumbers-in-action/auto-merge) - to merge approved PRs
* [issue-commentator](https://github.com/redhat-plumbers-in-action/issue-commentator) - to comment on PR with status of PR and merge status

Here are some examples of use from the systemd source-git. Please notice the `Pull Request validation` and `Auto Merge` sections:

* https://github.com/redhat-plumbers/systemd-rhel9/pull/281#issuecomment-2165555171
* https://github.com/redhat-plumbers/systemd-rhel9/pull/277#issuecomment-2139901121
* https://github.com/jamacku/nginx-container/pull/5#issuecomment-2182743930
Similar PR is here: https://github.com/sclorg/nginx-container/pull/306
/cc @jamacku

### DIFF
--- a/.github/auto-merge.yml
+++ b/.github/auto-merge.yml
@@ -1,0 +1,1 @@
+target-branch': ['master']
--- a/.github/workflows/auto-merge-on-demand.yml
+++ b/.github/workflows/auto-merge-on-demand.yml
@@ -1,0 +1,95 @@
+name: Auto Merge On Demand
+on:
+  schedule:
+    # Workflow runs every 45 minutes
+    - cron: '*/45 * * * *'
+  workflow_dispatch:
+    inputs:
+      pr-number:
+        description: 'Pull Request number/s ; when not provided, the workflow will run for all open PRs'
+        required: true
+        default: '0'
+
+permissions:
+  contents: read
+
+jobs:
+  # Get all open PRs
+  gather-pull-requests:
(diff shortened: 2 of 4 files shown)


jamacku (commented):
LGTM, But I would suggest you to create a composite GitHub action for sclorg that would call:

* redhat-plumbers-in-action/gather-pull-request-metadata@v1
* redhat-plumbers-in-action/pull-request-validator@v2
* redhat-plumbers-in-action/auto-merge@v2
* redhat-plumbers-in-action/issue-commentator@v1

It should make the maintenance of automation easier since it should be in one place.
Here is an example for inspiration: https://github.com/redhat-plumbers-in-action/source-git-automation/blob/main/action.yml

phracek:
I agree, but let's test it on two containers and later on we will decide whether we will use it in all repositories.

PR #312, opened 2024-07-24
Do not build s2i-core and s2i-base for F38.

phracek:
Do not build s2i-core and s2i-base for F38.
They reached EOL


jamacku (approved):
LGTM

PR #316, opened 2024-09-18
Add io.openshift.tags label to Dockerfile.rhel9

ckelleyRH:
Downstream patch synced upstream.


ckelleyRH:
For future archaeologists - the labels added here are required to build with Konflux.

phracek:
[test]

phracek (approved):
LGTM. Do not break container functionality

PR #320, opened 2024-12-05
CentOS Stream 10 is available on quay.io/centos/centos:stream10

phracek:
CentOS Stream 10 is available on quay.io/centos/centos:stream10

Let's switch it.


phracek:
[test]

frenzymadness (approved):
LGTM

PR #362, opened 2026-04-28
Add F44 dockerfiles and update build and push workflow

pkhartsk:
Fedora 44 is GO and releases today, so we're safe to use it for our images.


pkhartsk:
[test-all]

frenzymadness (approved):
Looks good to me.

pkhartsk:
Tests for fedora fail because the images are only tagged correctly in github workflows/build and publish, so in these tests it can't find s2i-core:44 since that tag doesn't exist. I ran the tests locally and they pass.

